PR #3, opened 2014-05-31
Fix syntax error on build.py --python parameter evaluation

reingart:
It is missing a parenthesis that causes a syntaxis error when trying to build with other interpreter::

    reingart@s5ultra:~/src/wxPython_Phoenix-trunk$ python build.py --python=/usr/bin/python3
    Command '"/usr/bin/python3" -c "import sys; print(sys.version[:3]"' failed with exit code 1.
      File "<string>", line 1
        import sys; print(sys.version[:3]
                                        ^
    SyntaxError: unexpected EOF while parsing

### DIFF
--- a/build.py
+++ b/build.py
@@ -210,7 +210,7 @@ def setPythonVersion(args):
             else:
                 PYTHON = args[idx+1]
                 del args[idx:idx+2]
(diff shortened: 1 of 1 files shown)


RobinD42:
Merged by hand due to missing source branch

PR #8, opened 2015-03-17
Fixes for running on Ubuntu 14.04.2

davidfraser:
This allows building on Ubuntu 14.04.2 with less headache :)

One commit fixes compatibility with graphviz 2.36.0 (as on Ubuntu 14.04.2) - see [this message](https://groups.google.com/forum/#!topic/wxpython-dev/FdvtlA5Og60) for another user with this problem... Without this patch, `etg` fails with an error message about the `map` file not being present

The other commit checks that the tools downloaded are executable. The downloaded `doxygen-1.8.8-linux` fails to execute on my Ubuntu 14.04.2 x86; this at least tells you about this and how to fix it in advance

### DIFF
--- a/build.py
+++ b/build.py
@@ -488,6 +488,13 @@ def getTool(cmdName, version, MD5, envVar, platformBinary):
                 print('       expected          "%s"' % md5)
                 print('       Set %s in the environment to use a local build of %s instead' % (envVar, cmdName))
                 sys.exit(1)
+            try:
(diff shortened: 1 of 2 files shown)


RobinD42:
I'll go ahead and merge this as-is, but for bonus points you may want to check if the cmd is executable without actually trying to run it, (`os.path.isfile(cmd) and os.access(cmd, os.X_OK)` or similar) and try setting the execute permisisons if possible (`os.chmod`) before failing.

PR #11, opened 2015-03-18
Check file is a standard file (not a directory), and that there are read...

davidfraser:
... and execute permissions on it, before trying to check md5sum and actually executing it

In response to @RobinD42 commenting on merge request #8 (though with the platformBuild the logic has become slightly convoluted) ... in case this helps

### DIFF
--- a/build.py
+++ b/build.py
@@ -481,22 +481,27 @@ def getTool(cmdName, version, MD5, envVar, platformBinary):
 
         msg('Checking for %s...' % cmd)
         if os.path.exists(cmd):
-            m = hashlib.md5()
-            m.update(open(cmd, 'rb').read())
-            if m.hexdigest() != md5:
-                print('ERROR: MD5 mismatch, got "%s"' % m.hexdigest())
-                print('       expected          "%s"' % md5)
-                print('       Set %s in the environment to use a local build of %s instead' % (envVar, cmdName))
-                sys.exit(1)
-            if platformBinary:
-                try:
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -481,22 +481,27 @@ def getTool(cmdName, version, MD5, envVar, platformBinary):
...
-                    print('       Set %s in the environment to use a local build of %s instead' % (envVar, cmdName))
-                    sys.exit(1)
-            return cmd
+            if not os.path.isfile(cmd):
+                print('ERROR: %s is not a standard file' % cmd)
+            elif not os.access(cmd, os.R_OK | os.X_OK):
+                print('ERROR: Cannot execute %s: permissions error' % cmd)
+            elif not platformBinary:
+                return cmd
+            else:
```
RobinD42: It shoudl still do the MD5 check even if it is a platform-specific binary.  What doesn't need to be done is to try and execute it direclty because it will not be a native executable.  Currently the waf tool is the only one that falls into this category and it is executed with python.

RobinD42:
Let's also change the order of tests a little.  This seems to make a little more sense to me:

isfile --> md5 check --> if platformBinary: return else --> os.access --> test executing

Also notice the change pushed today allowing the md5 check to be skipped, that should be retained although it shouldn't just return immediately like it does now, but continue evaluating the chain above.

RobinD42:
I had some time tonight and this was on my mind so I went ahead and implemented it.  See dd1d9d580f0c644411084455ba1c638c9ad56e33

davidfraser:
Thanks, that's much cleaner

PR #13, opened 2015-03-27
Update docs to point to Phoenix github instead of svn

driscollis:


wernerfb:
Mike, may I suggest that you change some of the other "SVN" reference at the same time?

Definitely:
DocstringsGuidelines.rst

Not sure:
agw.__init__.py and other .py files referring to SVN

driscollis:
Werner,

I got the docstrings one done. I suppose I would have to checkout the whole branch and grep for the others? I don't see a good way to search the branch through github itself.

Mike

PR #15, opened 2015-03-28
avoid unicode error

topic2k:
changed str() to unicode() to fix unicode errors.


wernerfb:
unicode() is not present in Py3, maybe use six.text_type

topic2k:
Haven't used Python 3 yet. Changed it as you suggested and at least it works on Python 2.7.9. Don't have the time right now to do more with Python 3, holiday is near :)

PR #16, opened 2015-03-30
fix speedmeter so demo of it does not need CallLater

wernerfb:


wernerfb:
keep it all in #101 in Robin's repo

PR #17, opened 2015-04-08
clearified the build commands

buhtz:
Modified the README.rst to make it more clear that you have to run all four build commands to build the stuff.

And this PR is kind of a test if I do PRs the right way. Do I? ;)


RobinD42:
I've made a similar change.

buhtz:
What is about the PR itself?

Is this PR in technical parameters correct and you could accept it? Did I used git the right way?

RobinD42:
The PR was fine.  I just didn't merge it because I rewrote the text a little differently.

PR #18, opened 2015-05-01
Fixed build errors on Windows.

calconst:
- The build process now uses regen.bat on Windows instead of the regen.sh which caused errors.
- Added missing prerequisites to the Windows build section in the main readme file.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,7 +33,7 @@ provided by the build.py script. There is also a setup.py script available
 for those who are used to the standard distutils or setuptools types of
 builds. The setup.py script assumes that all of the code generation steps
 have already been performed, and so it is suitable for use when building from
-a source snapshot tarball or when using easy_install or pip. The seup.py
+a source snapshot tarball or when using easy_install or pip. The setup.py
 script will delegate to build.py for the actual build, and build.py will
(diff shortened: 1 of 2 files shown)


DietmarSchwertberger:
With revision 1830, this modification does not work for me. Its probably about mixing directory separators / and \.
What's working, is to stick with ext\wxWidgets\docs\doxygen\regen.sh, but fix line endings in this script from \r\n to \n.

DietmarSchwertberger:
OK, I had another look.
regen.bat calls doxygen under exactly this name, while build.py downloads and stores doxygen as bin/doxygen-1.8.8-win32.exe
In build.py _doDox the variable doxCmd contains the full path to the doxygen command.
Maybe, it would be the best idea to move the functionality from regen.bat into _doDox and then call doxygen with the full name.
Or maybe it would be an option to make a regen_out.bat where the doxygen command is replaced with the full path.

When updating this pull-request, I would suggest to modify wscript as well to get rid of the touch command from cygwin:
 -   #task.exec_command('touch %s' % tgt)
 +   open(tgt, "wb").close()

Regards,
Dietmar

RobinD42:
This has been dealt with differently.  Closing.

PR #21, opened 2015-06-27
Implement SetExtraControlCreator for wx.FileDialog

conradchuang:
The Python callable passed to SetExtraControlCreator is saved as an attribute ("_wxExtraControlCreatorFunction") of the wxWindow wrapper.  This greatly simplifies Python object reference counting but does reserve a special attribute name.  The alternative of implementing reference counting in C is more cumbersome but is a possibility if reserving the attribute name is unacceptable.

### DIFF
--- a/etg/filedlg.py
+++ b/etg/filedlg.py
@@ -35,13 +35,64 @@ def run():
     isinstance(c, etgtools.ClassDef)
     module.addGlobalStr('wxFileDialogNameStr', c)
     module.find('wxFileSelectorDefaultWildcardStr').ignore()
-    
-    
-    # TODO: add this back. We'll need a way to pass it a callable that can be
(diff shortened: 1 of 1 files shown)


RobinD42:
It would be nice to have a unittest for this that tested (at least) that the new method exists and that the reference count is managed correctly.

RobinD42:
I've been working on this (finally!) and unfortunately I have discovered that it's possible for the creator function to be called with some other dialog (at least on Windows, in order to measure the widget it will create). That means that storing the creator function's PyObject can not be done by saving it as an attribute of the file dialog... I'll need to think about this some more in order to find another way to store and fetch the creator that is safe.

I'll make a new PR later with the changes I've made on top of this one.

PR #24, opened 2015-07-25
Use maxsize instead of maxint

balshetzer:
Python 3 no longer has sys.maxint

### DIFF
--- a/wx/lib/intctrl.py
+++ b/wx/lib/intctrl.py
@@ -43,7 +43,7 @@ import  wx
 
 #----------------------------------------------------------------------------
 
-from sys import maxint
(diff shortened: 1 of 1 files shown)


balshetzer:
It turns out there are other reasons why this module doesn't work with python 3.

wernerfb:
I started working on this yesterday, I think I have the Py3 issues done, but wanted to update the documentation and create a unittest before I do a PR.

balshetzer:
Thanks for tackling this!

wernerfb:
No problem, see #25

RobinD42:
Superseded by #25

PR #25, opened 2015-07-25
wx.lib.intctrl

wernerfb:
- update for Py3, is the use of sys.maxsize o.k.?
- update documenation
- create a simple unittest

### DIFF
--- a/wx/lib/intctrl.py
+++ b/wx/lib/intctrl.py
@@ -4,7 +4,7 @@
 # Created:      01/16/2003
 # Copyright:   (c) 2003 by Will Sadkin
 # License:     wxWindows license
-# Tags:         phoenix-port
+# Tags:        phoenix-port, py3-port
 #----------------------------------------------------------------------------
 # NOTE:
 #   This was written to provide a standard integer edit control for wxPython.
@@ -40,12 +40,17 @@ import  string
 import  types
 
 import  wx
+import wx.lib.six as six
 
 #----------------------------------------------------------------------------
 
-from sys import maxint
-MAXINT = maxint     # (constants should be in upper case)
-MININT = -maxint-1
+MAXINT = six.MAXSIZE     # (constants should be in upper case)
+MININT = -six.MAXSIZE-1
+
+if six.PY2:
+    LONGTYPE = long
+else:
+    LONGTYPE = int
 
 #----------------------------------------------------------------------------
 
@@ -648,7 +653,7 @@ class IntCtrl(wx.TextCtrl):
             value = self.GetValue()
 
         if( not (value is None and self.IsNoneAllowed())
-            and type(value) not in (types.IntType, types.LongType) ):
+            and type(value) not in six.integer_types ):
             raise ValueError (
                 'IntCtrl requires integer values, passed %s'% repr(value) )
 
@@ -693,7 +698,7 @@ class IntCtrl(wx.TextCtrl):
         for positive and negative values, respectively.
         """
         current_value = self.GetValue()
-        if not allow_long and type(current_value) is types.LongType:
+        if not allow_long and type(current_value) is LONGTYPE:
             if current_value > 0:
                 self.SetValue(MAXINT)
             else:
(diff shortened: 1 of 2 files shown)


gungad:
Apologies if this is the wrong place to report this but there is also an issue with `def _toGUI()` and possibly elsewhere with the same issue.
The issue: 'module' object has no attribute 'LongType', because as far as my limited knowledge allows me. I believe that there is only `int` now.
Full call-back error:
  File "/usr/local/lib/python3.4/dist-packages/wx/lib/intctrl.py", line 448, in __init__
    self, parent, id, self._toGUI(0),
  File "/usr/local/lib/python3.4/dist-packages/wx/lib/intctrl.py", line 748, in _toGUI
    elif type(value) == types.LongType and not self.IsLongAllowed():
AttributeError: 'module' object has no attribute 'LongType'

gungad:
I have knocked up a version of intctrl.py to cope with the LongType problem
Here is the 'diff' between my new version and the delivered copy.
I cannot attach my version of intctrl.py it seems that the web page doesn't like it, if the 'diff' is not enough, let me know and I'll find another way to send the code.
Sorry it's a mess, not sure how to format it.

diff intctrl.py intctrl.py.orig
<code>
12c12
< #   IntCtrl permits integer values to be retrieved or  set via
---
> #   IntCtrl permits integer (long) values to be retrieved or  set via
39a40
> import  types
44a46,51
> from sys import maxsize
> MAXINT = maxsize    # (constants should be in upper case)
> MININT = -maxsize -1
> 
> #----------------------------------------------------------------------------
> 
364a372
>          allow_long = False,
377c385,387
< 
---
>         If min < -sys.maxint-1 and the control is configured to not allow long
>         values, the minimum bound will still be set to the long value, but
>         the implicit bound will be -sys.maxint-1.
381a392,394
>         if max > sys.maxint and the control is configured to not allow long
>         values, the maximum bound will still be set to the long value, but
>         the implicit bound will be sys.maxint.
392a406,409
>     allow_long
>         Boolean indicating whether or not the control is allowed to hold
>         and return a long as well as an int.
> 
413c430
<                 limited = 0, allow_none = 0,
---
>                 limited = 0, allow_none = 0, allow_long = 0,
423a441
>         self.__allow_long = 0
441a460
>         self.SetLongAllowed(allow_long)
469c488
<         Returns the current integer value of the control.
---
>         Returns the current integer (long) value of the control.
505a525,526
>         If min > -sys.maxint-1 and the control is configured to not allow longs,
>         the function will return 0, and the min will not be set.
541a563,564
>         If max > sys.maxint and the control is configured to not allow longs,
>         the function will return 0, and the max will not be set.
628c651
<             and type(value) != int):
---
>             and type(value) not in (types.IntType, types.LongType) ):
662a686,707
>     def SetLongAllowed(self, allow_long):
>         """
>         Change the behavior of the validation code, allowing control
>         to have a long value or not, as appropriate.  If the value
>         of the control is currently long, and allow_long is 0, the
>         value of the control will be adjusted to fall within the
>         size of an integer type, at either the sys.maxint or -sys.maxint-1,
>         for positive and negative values, respectively.
>         """
>         current_value = self.GetValue()
>         if not allow_long and type(current_value) is types.LongType:
>             if current_value > 0:
>                 self.SetValue(MAXINT)
>             else:
>                 self.SetValue(MININT)
>         self.__allow_long = allow_long
> 
> 
>     def IsLongAllowed(self):
>         return self.__allow_long
> 
> 
703c748,751
<         elif type(value) != int:
---
>         elif type(value) == types.LongType and not self.IsLongAllowed():
>             raise ValueError (
>                 'IntCtrl requires integer value, passed long' )
>         elif type(value) not in (types.IntType, types.LongType):
740c788,791
<                 raise
---
>                 if self.IsLongAllowed():
>                     return long( value )
>                 else:
>                     raise
</code>

manns:
@gungad Could you e-mail me your intctrl.py file so that I can make a pull request to wernerfb?
My e-mail adddress is mmanns@gmx.net.

wernerfb:
In the PR types.LongType is replaced with LONGTYPE, which is "long" on Py2 and "int" on Py3.

wx/lib/intctrl.py
```diff
@@ -36,16 +36,22 @@
...
 
 import  wx
+import wx.lib.six as six
 
 #----------------------------------------------------------------------------
 
-from sys import maxint
-MAXINT = maxint     # (constants should be in upper case)
-MININT = -maxint-1
+MAXSIZE = sys.maxsize     # (constants should be in upper case)
```
RobinD42: The six module has a MAXSIZE constant.

wx/lib/intctrl.py
```diff
@@ -36,16 +36,22 @@
...
+MINSIXE = -sys.maxsize-1
```
RobinD42: Spelling error: MINSIXE --> MINSIZE

PR #33, opened 2015-09-13
- Colour.SetFromString replaced in Phoenix

wernerfb:


Metallicow:
@wernerfb I noticed a few days ago that HideNativeCaret is missing from the table also.
http://xoomer.virgilio.it/infinity77/wxPython/Widgets/wx.TextCtrl.html#HideNativeCaret
Could you do a pull reqz for it also. It should be included in the table I think...

wernerfb:
Hi,

Is that maybe an oversight?

http://trac.wxwidgets.org/ticket/5990
https://github.com/wxWidgets/wxWidgets/blob/master/include/wx/msw/textctrl.h

Looks like that is another thing for Robin's todo list.

Werner

Metallicow:
Maybe, but I wouldn't think so, as that ticket is 12 years old...
@vadz Could you clear this up/answer: ...About HideNativeCaret missing or removed in wxWidgets/wxPython? We are on the python end of things.

vadz:
Sorry, no idea what is this about, `HideNativeCaret()` was, ever since its addition [back in 2003](https://github.com/wxWidgets/wxWidgets/commit/e3a6a6b2bc15ddda1d614dd232d37e59ad024044) wxMSW-only and it's still present in wxMSW AFAICS.

RobinD42:
Since the native caret methods are not documented they don't get automatically added to Phoenix.  Since they are MSW-only they need special treatment anyway, and probably should not be in the docs.  I'll add them to Phoenix shortly...

PR #34, opened 2015-09-16
Fix some Py3 issues

wernerfb:
Some changes done to SVN which did not make it to Phoenix.


ChrisBarker-NOAA:
Looks good to me -- thanks!

I  don't have permission to merge, though.

wernerfb:
Robin will get to it when he finds time:)

wernerfb:
Tested this with the FC demo on Py2.7 and Py3.4 and with my own stuff but only on Windows.

PythonCHB:
Thanks!

PR #38, opened 2016-03-01
Two small fixes to build process on Windows

tibogens:
Hi Robin,
I have successfully compiled Phoenix with Python 3.4.4 on Windows - it's great to see Phoenix getting mature.
I had to apply two patches to make this happen. Although my fixes are not perfect, they at least allow the build to run out of the box.
Cheers
Thibault

### DIFF
--- a/build.py
+++ b/build.py
@@ -85,6 +85,14 @@ doxygenMD5 = {
 # And the location where they can be downloaded from
 toolsURL = 'http://wxpython.org/Phoenix/tools'
 
+# Cygwin is needed on Windows. This tool assumes it is installed in
+# the default location. Try to at least guess if we are using 32 bit
+# or 64 bit builds.
+if isWindows:
+    cygwin_path_32 = 'c:/cygwin'
+    cygwin_path_64 = 'c:/cygwin64'
+    cygwin_path = cygwin_path_64 if os.path.isdir(cygwin_path_64) else cygwin_path_32
+
 #---------------------------------------------------------------------------
 
 def usage():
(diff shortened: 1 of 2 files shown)


cztomczak:
I don't see cygwin being mentioned in README's build instructions.

tibogens:
Neither did I but it is used by the `build.py dox` step. It is the first time I build Phoenix so I have no idea if this step is actually needed (I just followed the main readme.rst).

tibogens:
OTOH MinGW is mentionned but not usable right now, so it might just be that Cygwin was simply overlooked as a requirement. (as many of us do have a copy of Cygwin lying around in C:\cygwin for all these funny scripts that require a bash interpreter ... except if you have 64-bit as I do)

DietmarSchwertberger:
Well, maybe it would be better to get rid of the Cygwin dependency for "user" builds, as this always causes some confusion.
Some weeks ago I tried the attached implementation of doDox. I did not try to build the documentation, though.
Should I submit a pull request?

[build_doDox.txt](https://github.com/wxWidgets/Phoenix/files/153413/build_doDox.txt)

Regards, Dietmar

RobinD42:
@DietmarSchwertberger: I'd rather not try to keep something like that in sync if there are changes to the original in wxWidgets sometime in the future. (Which there probably will be as soon as they change which doxygen is being used or new options are needed.)

Personally I don't think that requiring cygwin or similar is too onerous. Compared to the complexity of everything else in the build, cygwin is fairly easy.  I thought I had made a note of it in the README, but I guess not.  I'll add that.

Another idea that comes to mind is that the user most likely has git installed (unless they are building from the tarball, in which case all the generated files are already there, so no need to run dox.)  And with git on windows you also get Git Bash.  We could look for that and use it if cygwin isn't there.

DietmarSchwertberger:
I'm fine with either solution (add Cygwin64 detection, the Python alternative or even the not-yet-tried alternative of calling regen.bat after defining an alias using DOSKEY).

PR #68, opened 2016-04-18
Fix wacky ints (mainly for Py2.7 / Win64)

swt2c:
The 'wacky ints' tests were failing on Win64.  It seems that the PyLong_XX
functions don't work correctly there on Py 2.7.  The solution was to switch to
the PyInt_XX functions, which will correctly return a PyLong when necessary.

### DIFF
--- a/src/wacky_ints.sip
+++ b/src/wacky_ints.sip
@@ -20,7 +20,7 @@
 %MappedType size_t {
 
     %TypeHeaderCode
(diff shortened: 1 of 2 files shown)


swt2c:
Actually, hold off on looking at this one just yet.  I have an idea for a slightly better fix.

swt2c:
Updated.

RobinD42:
Thanks!

PR #92, opened 2016-05-27
Fix ownership transfer for wxXmlDocument methods

swt2c:


swt2c:
Of course, this does not prevent someone from reusing objects that have already been transferred over to the C++ side, for example this, which will crash:

```
rootNode = wx.xml.XmlNode(wx.xml.XML_ELEMENT_NODE, "TestXmlFile")
doc = wx.xml.XmlDocument()
doc.SetRoot(rootNode)
doc1 = wx.xml.XmlDocument()
doc1.SetRoot(rootNode)
```

I guess we are just expecting people to read the wxWidgets documentation and not do that?  The only other thing I could think we could do is make a copy of the Python object automatically behind the scenes, but this would add some extra overhead.

PR #96, opened 2016-06-01
Remove spurious backticks from GridBagSizer demo

GadgetSteve:
A couple of the print statements in GridBagSizer demo had gained spurious backticks which prevent it running - removed these and it runs fine.


GadgetSteve:
Also added fixes for demos of ColourDB and DrawXXXList.

demo/DrawXXXList.py
```diff
@@ -4,6 +4,7 @@
 import time
 
 import wx
+import wx.lib.six as six
```
RobinD42: six isn't used in the module so there is no need to add an import for it.

swt2c:
Also, can you get rid of those merge commits?

GadgetSteve:
@RobinD42 - No problem removing the six import & adding repr however I notice an exception when clicking on the two "Move this to (3,6)" buttons of:

    AttributeError: 'GridBagSizer' object has no attribute 'CheckForIntersectionPos'

Since the demo is also largely a test suit I figure that I shouldn't try to fix this as it is correctly highlighting a missing method.

@swt2c I don't know how to keep my fork in step with the master without the merge commits. If you have any hints ...

RobinD42:
> no attribute 'CheckForIntersectionPos'

Thanks, I'll take care of that.

> I don't know how to keep my fork in step with the master without the merge commits. If you have any hints ...

Merge Phoenix master into your master as needed, but you should always make your changes on a branch, one branch per PR that you plan on submitting.

PR #97, opened 2016-06-04
Remove shebang from non-executable files - fixes rpmlint errors

swt2c:


RobinD42:
Thanks.  Those keep sneaking back in from time to time. 😄

Metallicow:
Maybe writing a small snippet of a 'code quality' script(s) to occasionally run could check for this bit...
I believe a "Code Quality" suite is just as important as a "Unit Test" suite.

> Now is better than never.
> Although never is often better than *right* now.
> If the implementation is hard to explain, it's a bad idea.
> If the implementation is easy to explain, it may be a good idea.

PR #98, opened 2016-06-10
Additional Refactoring of lib.plot [Part 1]

dougthor42:
This PR is a general cleanup and refactoring of lib.plot.

Items in this PR:
+ [x] Remove duplicate code found in `PolyBars` and `PolyHistogram`, possibly by one a special case of the other. TBD.
+ [x] Add `SaveBrush` decorator and `SavedBrush` context manager
+ [x] Add general `TempStyle` combination decorator and context manager
+ [x] update `PolyPoints.points` property - move log10 logic to `self._log10`
+ [x] remove `eval` in `PolyMarkers._drawmarkers` (unsafe)
+ [x] refactor `PlotCanvas.EnableAxes`, `EnableAxesVales` and `EnableTicks` - they all currently have the same logic which could be pulled out to a separate function
+ [x] ~~Change `PlotCanvas._drawTicks` to use the new `XSpec` and `YSpec` API.  *Needs investigation - Not quite sure what I meant here*~~ Already implemented.
+ [x] ~~Add option to individually enable or disable specific axes.~~ Already implemented.
+ [x] `PlotCanvas._drawAxesValues`: replace dc.DrawText with dc.DrawTextList?
+ [x] Clean up math in `PlotCanvas._drawPlotTitle`
+ [x] `PlotCanvas._drawAxesLabels`: Clean up math
+ [x] `PlotCanvas._drawPlotAreaLabels`: Clean up call signature
+ [x] follow PEP8 guidelines on max line length
+ [x] Documentation
+ [x] verify how `NaN` is handled in `BoxPlot.calcBpData` and document it.

To be saved for another PR:
+ ~~[ ] Add much more customization options (pens, brushes) to `PolyBars` and `PolyHistogram`~~
+ ~~[ ] convert `PlotGraphics.LogScale` to `try..except`~~
+ ~~[ ] rename items in `PlotCanvas.EnableDiagonals` to something else. Or perhaps switch to enums?~~
+ ~~[ ] remove deep copy in `PolyPoints.points` property~~
+ ~~[ ] clean up large `if` statement and math in `PlotCanvas._Draw`.~~
+ ~~[ ] possibly able to modify event handlers in `PlotCanvas.UpdatePointLabel` so that the event is only bound if the item is enabled. This would remove many of the if statements within the event handlers.~~
+ ~~[ ] Add an option to draw ticks outside to plot rather than only inside.~~
+ ~~[ ] I have a note that `PlotCanvas._drawAxesValues` has potential code duplication (with `_drawGrid` and `_drawTicks`)~~
+ ~~[ ] `PlotCanvas._drawAxesValues`: update the bounding boxes when adding right and top values~~
+ ~~[ ] `PlotCanvas._drawAxesLabels`: Fix bug where axes values get too big when "this" is turned off. *Not sure what "this" is...*~~
+ ~~[ ] **Create unit tests** (I can't believe I forgot this one! I should be ashamed...)~~
+ ~~[ ] Update pointLabels to be easier to read~~
  + ~~add semi-opaque background~~
  + ~~adjust label location if it would fall out of bounds.~~
+ ~~[ ] Adjust location of changelog? It's not pythonic to have a large block of comments at the top of a module.~~

### DIFF
--- a/unittests/test_lib_plot.py
+++ b/unittests/test_lib_plot.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+import unittest
+import wtc
+import wx
+
+import wx.lib.plot as wxplot
+
+#---------------------------------------------------------------------------
+
+class lib_plot_PlotCanvas_Tests(wtc.WidgetTestCase):
+
+    def test_lib_plot_plotcanvasCtor(self):
+        """ Ctor? """
+        p = wxplot.PlotCanvas(self.frame)
+
+
+class lib_plot_Tests(wtc.WidgetTestCase):
+    def test_lib_plot_tempstyle_contextmanager(self):
+        pass
+
+    def test_lib_plot_tempstyle_decorator(self):
+        pass
+
+#---------------------------------------------------------------------------
+
+if __name__ == '__main__':
+    unittest.main()
--- a/wx/lib/plot.py
+++ b/wx/lib/plot.py
@@ -497,7 +497,7 @@ class PendingDeprecation(object):
         `old_func` is pending deprecation. Please use new_func instead.
 
     """
-    _warn_txt = "`{}` is pending deprecation. Please use {} instead."
+    _warn_txt = "`{}` is pending deprecation. Please use `{}` instead."
 
     def __init__(self, new_func):
         self.new_func = new_func
@@ -706,13 +706,11 @@ class PolyPoints(object):
                                                     # TODO: get rid of the
                                                     # need for copy
 
-        # TODO: I can make this better... move logic to self._log10
         # work on X:
         if self.AbsScale[0]:
-            data[:, 0] = np.abs(data[:, 0])
+            data = self._abs(data, 0)
         if self.LogScale[0]:
-            data = np.compress(data[:, 0] > 0, data, 0)
-            data[:, 0] = np.log10(data[:, 0])
+            data = self._log10(data, 0)
 
         if self.SymLogScale[0]:
             # TODO: implement SymLogScale
@@ -723,10 +721,9 @@ class PolyPoints(object):
 
         # work on Y:
         if self.AbsScale[1]:
-            data[:, 1] = np.abs(data[:, 1])
+            data = self._abs(data, 1)
         if self.LogScale[1]:
-            data = np.compress(data[:, 1] > 0, data, 0)
-            data[:, 1] = np.log10(data[:, 1])
+            data = self._log10(data, 1)
 
         if self.SymLogScale[1]:
             # TODO: implement SymLogScale
@@ -742,6 +739,11 @@ class PolyPoints(object):
         """ Take the Log10 of the data, dropping any negative values """
         data = np.compress(data[:, index] > 0, data, 0)
         data[:, index] = np.log10(data[:, index])
+        return data
+
+    def _abs(self, data, index):
+        """ Take the Abs of the data """
+        data[:, index] = np.abs(data[:, index])
         return data
 
     def boundingBox(self):
@@ -2178,7 +2180,7 @@ class PlotCanvas(wx.Panel):
         frame.Centre(wx.BOTH)
         frame.Show(True)
 
-    @PendingDeprecation("self.LogScale")
+    @PendingDeprecation("self.LogScale property")
     def setLogScale(self, logscale):
         self.LogScale = logscale
 
--- a/wx/lib/plot.py
+++ b/wx/lib/plot.py
@@ -198,36 +198,36 @@ class TempStyle(object):
     Combination Decorator and Context Manager to revert pen or brush changes
     after a method call or block finish.
 
-    **Note: Can only be used on class methods.**
+    :param which: The item to save and revert after execution. Can be
+                  one of ``{'both', 'pen', 'brush'}``.
+    :type which: str
+    :param dc: The DC to get brush/pen info from.
+    :type dc: :class:`wx.DC`
 
-    Parameters:
-    -----------
-    which : {'both', 'pen', 'brush'}
-        The item to save and revert after execution.
-    dc : wx.DC object
-        The DC to pull brush/pen information from.
+    ::
 
-    Usage:
-    ------
-    As a method decorator::
-
+        # Using as a method decorator:
         @TempStyle()                        # same as @TempStyle('both')
         def func(self, dc, a, b, c):        # dc must be 1st arg (beside self)
             # edit pen and brush here
 
-    As a context manager::
-
+        # Or as a context manager:
         with TempStyle('both', dc):
             # do stuff
 
-    Notes:
-    ------
-    As of 2016-06-15, this can only be used as a decorator for **class
-    methods**, not standard functions. There is a plan to try and remove
-    this restriction, but I don't know when that will happen...
+    .. Note::
 
-    Combination Decorator and Context Manager! Also makes Julienne fries!
-    Will not break! Will not... It broke!
+       As of 2016-06-15, this can only be used as a decorator for **class
+       methods**, not standard functions. There is a plan to try and remove
+       this restriction, but I don't know when that will happen...
+
+
+    .. epigraph::
+
+       *Combination Decorator and Context Manager! Also makes Julienne fries!
+       Will not break! Will not... It broke!*
+
+       -- The Genie
     """
     _valid_types = {'both', 'pen', 'brush'}
     _err_str = (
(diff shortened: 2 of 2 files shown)


RobinD42:
Sounds adventurous! 😃   I'm looking forward to seeing this come together.  Thanks Doug.

dougthor42:
@RobinD42 In unit tests, what does "Ctor" mean?

Example:
```python
class lib_floatcanvas_floatcanvas_Tests(wtc.WidgetTestCase):

    def test_lib_floatcanvas_floatcanvasCtor(self):
        fccanvas = fc.FloatCanvas(self.frame)

    def test_lib_floatcanvas_navcanvasCtor(self):
        self.navcanvas = nc.NavCanvas(self.frame)
```

swt2c:
> @RobinD42 In unit tests, what does "Ctor" mean?

Constructor

RobinD42:
Correct. It's an abbreviation commonly used in C++.  In the unit tests and elsewhere I use it to mean "creating the object" so, in other words, in tests it would imply testing the `__init__` method.

dougthor42:
Ah, C++ stuff. Thanks for the info.

Do you have any special requirements for the unit tests?

I'm trying to keep the naming and docstring conventions the same as most of the other code, but other than that I'll probably just do my own thing unless you say otherwise.

RobinD42:
I haven't set any requirements for tests other than naming conventions which you've already figured out. I tend to first add tests that simply create instances of the important of complex classes in the module, then I try to add some test that  populate or otherwise exercise the classes like how they would typically be used, and then some tests that verify critical parts of the code or those that may be fragile, if any.  

For the extension modules I also tend to add tests that check for the existence of module level constants, event binders, etc. since it is easy for those to disappear without notice until some user code tries to use it.  That is less of an issue with python modules though so I don't mind if that is done or not with them.

dougthor42:
A couple other questions:

1. Should I target Py3 only? Or should I maintain compatibility with 2.7?
  + I haven't been able to find any doc stating whether or not Phoenix will support 2.7.
  + If only Py3, what min version? 3.3?
2. How stable should I keep the API?
  + There are a lot of deprecated methods (primarily getters and setters) that could be removed. These were deprecated on Feb 27, 2016 in #26. When should they be removed?
  + I'd like to clean up some naming conventions, switching from `CamelCase` to `mixedCase` for properties. Should this be done with a deprecation timeline (how long?) or should I just go ahead and make the change?
3. I plan on taking this module and converting it to a subpackage in a later PR - is anyone opposed to that?
  + Includes: moving demo to separate file, creating standalone changelog so that it's not cluttering up the module, moving out utility functions, and other actions TBD.

psionprime:
I'm not a developer but subscribe to the list as I've recently been 
doing heavy development with wxPhoenix. My 2 cents follow as when I 
started with my project, I mulled over this same question. I went with 
supporting 3.5.1 as it was the most current stable release and 2.7 is 
depreciated.

When confronted with this type of question and a new project, I go with 
the latest and greatest then support it for 5 years. If 2.7 is kept 
alive with backports it will eventually become so close to 3 as to have 
little impact. I've been using wxFormBuilder to kick start my GUI work 
and it still produces 2.7 code and wxPython but the changes are minimal 
to get it to 3.5. A con is 3.5 is slower than 2.7 but not much.

Again, not a dev, just a guy giving feedback who appreciates the 
excellent efforts of this group !

RobinD42:
@dougthor42: 

1. Phoenix will be supporting Python 2.7 while it is still commonly used, so `wx.lib.plot` will need to do so also.  You can use the `wx.lib.six` module to help smooth over many of the differences, as well as `from __future__ import whatever`. For Python3 we're following the standard of what was done in the past with Classic and keeping compatibility with the current and previous Python release series.  So currently that is Python 3.5.x and 3.4.x.

2. Maintaining API stability can be a tricky issue. Some things in Phoenix core are being intentionally broken as part of the whole Phoenix philosophy, but there has to be a very good reason for it because each and every one will need to be justified to irate users at some point in the future. So my preference is to try to keep as much API stability as possible (if not all of it), but if you want to start migrating things to a new API style or whatever then I don't mind if the old names raise `DeprecationWarnings` (I've done a fair bit of that in the core.)  I think that there needs to be at least one official release when the old names are still available but deprecated, before they are removed.

3. Yes, a subpackage sounds like a good idea, as long as compatibility is maintained as described in 2.

dougthor42:
Sounds great. I'll target 2.7 and 3.4+, keep the API stable with raised warnings, and will submit another PR for converting to a subpackage once this one is done.

Thanks,

dougthor42:
@RobinD42 

Oh, one last question for now: Should API that was added recently (#26), be considered stable? Or can it be changed because Phoenix is still in preview?

Basically, I added a whole bunch of new properties in #26 to deprecate the old `GetFontSizeAxis`/`SetFontSizeAxis` methods and would like to change the new properties to follow the rest of Phoenix's naming conventions.

RobinD42:
I think that changing things in the API that are new would be okay.

dougthor42:
@RobinD42 

I think this PR is getting far too big, so I'm going to split it into 2 or 3 (or possibly more) PRs. I've updated the original message to reflect what's in this one and what will be in other PRs.

Please review when you can. The next PR will be moving the module to a subpackage, as that will make it easier to work with.

Also please let me know if you want me to squash any commits.

RobinD42:
Github says the diff is too big to display all the changes, so I can't do inline comments.  So splitting into multiple PRs is a good call.  I'll add some general comments here instead.

* **TempStyle**: We already have some similar classes for saving/restoring the GDI elements on a DC, see the `wx.DC*Changer` classes, which can be used as a context manager.  `TempStyle` is more comprehensive and more capable with decorator support, so I don't mind keeping it, but perhaps since it is generally useful it would make sense to move it to the core library, name it similar to the others, (maybe `DCStyleChanger` or something) and add font and perhaps text color support.

* **SavePen, SavedPen, SaveBrush, SavedBrush**  Are you planning on replacing all uses of these with `TempStyle`? It looks like from the rest of the code diffs that you are, so these could probably just be removed from the module. If not then at least removing the context managers and replacing them with the appropriate `wx.DC*Changer` class would be a good refactor.  I don't think anybody would be using these outside of the plot module, but maybe for now they could be replaced with a simple deprecated stub that is implemented with `TempStyle`, just in case.

* **PolyPoints.scaleAndShift**: I think you are correct that `!=` is what should be used.  It shouldn't matter if the same tuple objects are used there, just whether they are equal or not.

* A bit more description in the class docstrings would be nice. For example, "Creates PolyLIne object" tells me nothing about what it is and/or what it is used for, other than it probably has something to do with lines.  I think one of the typical classes of users of this module are those for whom matplotlib is to confusing or has too much overhead, so making it as simple and self explanatory as possible is a good idea. (And yes I realize that those descriptions were already there in the code before this PR, but since you are updating the docs anyway it would be good to make changes like this too.)  

* Another example of possible docstring improvements is with the various properties. I see a lot of "The current fooBar value." along with type info and etc., but that doesn't tell me anything about what the property does or is used for, unless the name is very descriptive.  Either way it would be good for the docstring to give enough info about what it is or what it is used for that the programmer doesn't have to guess or use trial/error to figure it out.  

Otherwise, it looks to me like you are taking this in the right direction.  Thanks for your work on this module (soon to be package)!

dougthor42:
Yeah, I noticed that GitHub said the diff was too big so I stopped.

I've done some of the changes, but others will be in different PRs.

+ **TempStyle:** I'll definitely take a look at the `wx.DC*Changer` classes - I didn't know about them before. I'll submit a separate PR to add the TempStyle (renamed) Decorator/CM to the core rather than being in this lib. It still needs more work though, it's not quite where I'd like it to be.
+ **SavePen, et. al.:** Yeah we can remove those - they've been replaced with TempStyle.
+ **PolyPoints.scaleAndShift:** Fixed. It looks like the reason that it was `is not` was because one of the items is a tuple/list, while the other is a `np.ndarray` object. Running `list != np.array` returns an array of booleans, whereas `is not` will always return `True`. 
+ **Class Docstrings:** Added to TODO for next PR
+ **Property Docstrings:** Yeah, I wasn't happy with it either, but I couldn't figure out a better way. I wasn't able to find many examples of property docstrings. I'll revisit it in the next PR.

dougthor42:
Awe crap, it looks like I messed up a rebase - this PR now has a bunch of extra commits in it. Sorry about that!

How would you like me to proceed?

dougthor42:
Ok I'm pretty sure I fixed it. Sorry again!

PR #100, opened 2016-06-11
Workaround multiple inheritance bug in SIP 4.18 with vscroll

swt2c:
In SIP 4.18, the cast_xyz function generation has been changed, but a bug was
introduced where it attempts to cast a subclass to a multiply inherited
superclass.  This doesn't compile due to ambiguity, so workaround the problem
by deleting the extra casts in the generated code.


RobinD42:
No longer needed due to updating to sip 4.18.1 (snapshot)

PR #114, opened 2016-07-08
Update helpviewer.py

david-hughes:
Trivial changes to comments - practice creating a pull request


RobinD42:
Hi David, I have just one comment about PR best practices.  It is better to make your local changes on a new branch instead of master, and push that branch for the PR.  Currently if you have other unrelated changes to make or you want to update from upstream, then those commits to your master would be added to this PR which is usually not what you want.

PR #117, opened 2016-07-11
Additional Refactoring of lib.plot [Part 3: other changes]

dougthor42:
## Update 2017-09-08

I will not be able to finish all the items in the PR 😢 

Per @RobinD42's suggestion, I have made issues for all unfinished items so that someone else can complete them.

This PR now only makes the following changes:
+ Add an option to draw ticks outside to plot rather than only inside.
+ Fix instances of `wrapper(*args, **kwargs)` in built docs
+ Updates docs
+ Addes additional "Ticks" demos
+ Miscellaneous cleanup

All the other items that I wanted to accomplish have their related Issue listed below.

+ #512
+ #513 
+ #514 
+ #515 
+ #516
+ #517 
+ #518
+ #519 

## Original PR message

(This PR replaces #112, as I mucked up the git history too much)

This PR will (hopefully) finish off all the changes that I had planned for lib.plot.

It will also include any feedback from Robin and anyone else.
- [ ] Add much more customization options (pens, brushes) to `PolyBars` and `PolyHistogram` (#512)
- ~~[ ] convert `PlotGraphics.LogScale` to `try..except`~~
- ~~[ ] rename items in `PlotCanvas.EnableDiagonals` to something else. Or perhaps switch to enums?~~
- [ ] remove deep copy in `PolyPoints.points` property (#513)
- [ ] clean up large `if` statement and math in `PlotCanvas._Draw`. (#514)
- [ ] possibly able to modify event handlers in `PlotCanvas.UpdatePointLabel` so that the event is only bound if the item is enabled. This would remove many of the if statements within the event handlers. (#515)
- [x] Add an option to draw ticks outside to plot rather than only inside.
- [ ] I have a note that `PlotCanvas._drawAxesValues` has potential code duplication (with `_drawGrid` and `_drawTicks`) (#516)
- [ ] `PlotCanvas._drawAxesValues`: update the bounding boxes when adding right and top values (#517)
- [ ] `PlotCanvas._drawAxesLabels`: Fix bug where axes values get too big when "this" is turned off. _Not sure what "this" is..._
- [ ] Update pointLabels to be easier to read (#518)
  - add semi-opaque background
  - adjust label location if it would fall out of bounds.
- [ ] More documentation updates based on Robin's comments in https://github.com/wxWidgets/Phoenix/pull/98#issuecomment-230351215
  - [x] Fix instances of `wrapper(*args, **kwargs)` in built docs [**needs confirmation**]

### DIFF
--- a/wx/lib/plot/__init__.py
+++ b/wx/lib/plot/__init__.py
@@ -2,10 +2,7 @@
 # pylint: disable=C0413
 #   C0413: Import should be placed at the top of the module
 """
-wx.lib.plot
-===========
-
-This is a simple plotting library for the wxPython Phoenix project.
+A simple plotting library for the wxPython Phoenix project.
 
 """
 __version__ = "0.0.1"
@@ -42,7 +39,7 @@ from .polyobjects import PlotGraphics
 from .polyobjects import PlotPrintout
 
 from .utils import TempStyle
-from .utils import PendingDeprecation
+from .utils import pendingDeprecation
 
 # For backwards compat.
 BoxPlot = PolyBoxPlot
--- a/wx/lib/plot/examples/demo.py
+++ b/wx/lib/plot/examples/demo.py
@@ -509,7 +509,8 @@ class PlotDemoMainFrame(wx.Frame):
                     'Enables the display of the axes values')
 
         submenu = wx.Menu()
-        submenu_items = ("Bottom", "Left", "Top", "Right")
+        submenu_items = ("Bottom", "Left", "Top", "Right",
+                         "Bottom+Left", "All")
         help_txt = "Enables {} ticks"
         self.ticksSubMenu = submenu
         for _i, item in enumerate(submenu_items, 2501):
@@ -524,6 +525,8 @@ class PlotDemoMainFrame(wx.Frame):
         self.Bind(wx.EVT_MENU, self.OnEnableTicksLeft, id=2502)
         self.Bind(wx.EVT_MENU, self.OnEnableTicksTop, id=2503)
         self.Bind(wx.EVT_MENU, self.OnEnableTicksRight, id=2504)
+        self.Bind(wx.EVT_MENU, self.OnEnableTicksBottomLeft, id=2505)
+        self.Bind(wx.EVT_MENU, self.OnEnableTicksAll, id=2506)
 
         menu.Append(250, 'Enable Ticks', submenu,
                     'Enables the display of the ticks')
@@ -829,8 +832,8 @@ class PlotDemoMainFrame(wx.Frame):
         self.ticksSubMenu.Check(2502, self.client.enableTicks[1])
         self.ticksSubMenu.Check(2503, self.client.enableTicks[2])
         self.ticksSubMenu.Check(2504, self.client.enableTicks[3])
-#        self.ticksSubMenu.Check(2505, all(self.client.enableTicks[:2]))
-#        self.axesSubMenu.Check(2506, all(self.client.enableTicks))
+        self.ticksSubMenu.Check(2505, all(self.client.enableTicks[:2]))
+        self.ticksSubMenu.Check(2506, all(self.client.enableTicks))
 
     def OnEnableTicksBottom(self, event):
         old = self.client.enableTicks
@@ -854,6 +857,17 @@ class PlotDemoMainFrame(wx.Frame):
         old = self.client.enableTicks
         self.client.enableTicks = (old[0], old[1],
                                    old[2], event.IsChecked())
+        self._checkOtherTicksMenuItems()
+
+    def OnEnableTicksBottomLeft(self, event):
+        checked = event.IsChecked()
+        old = self.client.enableTicks
+        self.client.enableTicks = (checked, checked, old[2], old[3])
+        self._checkOtherTicksMenuItems()
+
+    def OnEnableTicksAll(self, event):
+        checked = event.IsChecked()
+        self.client.enableTicks = tuple([checked] * 4)
         self._checkOtherTicksMenuItems()
 
     # -----------------------------------------------------------------------
--- a/wx/lib/plot/plotcanvas.py
+++ b/wx/lib/plot/plotcanvas.py
@@ -14,6 +14,8 @@ __docformat__ = "restructuredtext en"
 
 # Standard Library
 import sys
+import string as _string
+import time as _time
 
 # Third-Party
 import wx
@@ -21,9 +23,10 @@ import numpy as np
 
 # Package
 from .polyobjects import PlotPrintout
+from .polyobjects import PolyMarker, PolyLine, PolyBoxPlot
 from .utils import DisplaySide
 from .utils import set_displayside
-from .utils import PendingDeprecation
+from .utils import pendingDeprecation
 from .utils import TempStyle
 from .utils import scale_and_shift_point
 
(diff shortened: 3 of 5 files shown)


dougthor42:
@RobinD42 

I'm working on the docs for this, trying to fix the docs showing `wrapper(*args, **kwargs)` for decorated functions and also updating cross references so that they work correctly

Example:
image
+ https://wxpython.org/Phoenix/docs/html/wx.lib.plot.plotcanvas.PlotCanvas.html#wx.lib.plot.plotcanvas.PlotCanvas
+ `enableAntiAiasing` should be a link


**However, I'm having trouble building the docs locally** so I can't preview the changes to see if they are working. Should I just assume that the changes are OK and then check them after the buildbot releases a snapshot after the merge? Or do you have another option?

The issues I'm having:
+ no static assets. Fixed by downloading the most recent docs and copying the _static/images/inheritance folder from them into my working dir.
+ `python build.py etg` builds the main docs but nothing from `wx.lib`.
+ `python build.py sphinx` builds the main docs but nothing from `wx.lib`.
+ running `make html` from the `docs/sphinx/` director also doesn't built the `wx.lib` docs.

Am I missing something in my build flow?
1. Cloned my repo `dougthor42/Phoenix.git`
2. Checked out my branch `refactor-lib.plot-part3`
3. `git submodule update --init`
4. `python build.py dox`
5. `python build.py etg`

RobinD42:
I was looking at the issue with the docs for the deprecated methods a few days ago.  It might be due to the commented-out line `self._update_docs()` in `PendingDeprecation` but I didn't test that. BTW, there is a `wx.deprecated` decorator in the core namespace, (however I see now that it isn't included in the documentation! Shame on me.)  It's quite a bit more complex than your `PendingDeprecation` class as it is able to mark classes and properties as deprecated in addition to functions and methods, so if you want to stick with yours that is okay.

The ReST docs for `wx.lib` are generated with a separate build command, `python build.py wxlib`.  Since it actually imports the files as it is scanning them you need to be sure that there is a built version of Phoenix on the python path.  If it's an installed one rather than a built Phoenix in the workspace, then you may need to do some tricks to make sure it is importing `wx.lib` from your project workspace instead of the installed `wx.lib`.

The files in `_static/images/inheritance` are generated by the `dot` utility from graphviz.  Copying them into place is probably an acceptable workaround if you don't have that installed.

dougthor42:
Thanks for the info, it allowed me to get the doc builds running locally so I can play around and try to fix things.

----

Couple of notes for Future Me:

1. I had to remove the call to `demo.run_demo()` in `__main__.py`, as that file would get executed and the demo would pop up. I've gotta come up with another solution for this - perhaps we just don't allow users to call `plot` as a module (`python -m wx.lib.plot`)?
2. Some research [[1]](http://stackoverflow.com/a/13258947/1354930) suggested adding the call signature as the first line of the docstring on the decorated function. This does not appear to work.
3. [Overriding the call signature](http://stackoverflow.com/a/15693082/1354930) in the `wx.lib.plot.txt` sphinx file is not feasible.
4. Perhaps switching to a function-based decorator will work...
5. `@functools.wraps(func)` decorator should take care of copying `__doc__` and `__name__`.
6. What about using the [`decorator`](https://pypi.python.org/pypi/decorator) package?
7. Can I use `wx.deprecated` as a decorator? And would that fix things?
8. Doc build process:
  1. `git clean -d`
  2. Make test changes to live wx in `.../site-packages/wx/lib/plot`
  3. Copy `inheritance` folder to `docs/sphinx/_static/images` (Graphvis website is down - can't download)
  4. `python build.py wxlib`
  5. `python build.py sphinx`
  6. Check html files.

RobinD42:
> I had to remove the call to demo.run_demo() in __main__.py, as that file would get executed and the demo would pop up. I've gotta come up with another solution for this - perhaps we just don't allow users to call plot as a module (python -m wx.lib.plot)?

I had already changed that file in the master branch, see 128b28c8795a629df03f207dd94cbfba50b01f4f

> Can I use wx.deprecated as a decorator? And would that fix things?

Yes, and I think so.  There is also `wx.deprecatedMsg` for when you want to use it as a decorator and also have a custom message.  It's a little awkward, but works.  If you see any improvements that can be made there feel free to propose something.  The implementations are here: https://github.com/wxWidgets/Phoenix/blob/master/src/core_ex.py#L39

dougthor42:
### Decorated Functions + Sphinx
Doesn't look like any decorators will work. I tried switching to function-based decorators and tried using `wx.deprecatedMsg` to no avail (the method would not even be added to the docs).

So what I've decided to do is simply add a line to the start of every deprecated function which raises `PendingDeprecationWarning` instead of decorating them:

```python
def GetFontSizeTitle(self):
    """
    Get Title font size (default is 15 point)

    .. deprecated:: Feb 27, 2016

       Use the :attr:`~wx.lib.plot.plotcanvas.PlotCanvas.fontSizeTitle`
       property instead.
    """
    pendingDeprecation('fontSizeTitle')
    return self.fontSizeTitle
```

```python
def pendingDeprecation(new_func):
    warn_txt = "`{}` is pending deprecation. Please use `{}` instead."
    _warn(warn_txt.format(inspect.stack()[1][3], new_func),
          PendingDeprecationWarning)
```

The above renders correctly:

image


----

### Properties + Sphinx
The other doc question I had was about properties. Right now, properties are rendered by sphinx as:

image

However, I know that properties can be rendered and that the [documentation must be on the getter](http://stackoverflow.com/a/17522081/1354930). I'd like to have the full docstring rendered rather than just two "see X, Y" links which point to the very item that they're under. Is this something that we can change?

A simple docstring on the property would be:
```python
@property
def fontSizeTitle(self):
    """
    The current Title font size in points.

    Use this property to change the size of the plot title. All units are in pts, and the default size is 15pt.

    :getter: Return the size of the plot title.
    :setter: Change the size of the plot title.
    :type:   int or float
    """
    return self._fontSizeTitle
```

RobinD42:
**Decorated functions**: I thought that I had verified `wx.deprecated` was working as desired, but I may have looked at the wrong example or something.  Your workaround is fine with me.

**Properties**: Andrea's implementation for generating docs for Python code doesn't actually use Sphinx's module scanner tool, etc. It does all the introspection and docstring extraction itself, in order to be able to generate things the same for those modules as it does for the extension modules.  So the issue here is simply that the `wxlib` build command is not doing anything with the properties other than listing and linking the getter and setter methods. It would be great to have it be able to do more than that. If you're interested in working on enhancing that part of the docs generation, the code to tweak is in the `Property` class in `Phoenix/sphinxtools/librarydescription.py`.

RobinD42:
I took another look at it and realized that most of what was needed is already there, so I just pushed a change on master that will output the getter's docstring it is has one.

RobinD42:
@dougthor42, just FYI I made some minor changes to `plot/examples/demo.py` on master.  See ce00fc06a5f778860f6dbb4db2273b54a813e997

dougthor42:
Thanks for the info! Also thanks for making the getter docstring change

And sorry for stalling on this PR. Things have picked back up for me so I haven't had much time recently.

RobinD42:

dougthor42:
@RobinD42 It's been over a year, I'm so sorry! :sob: I don't think I'll be able to get around to finishing this.

The without this PR, the `plot` package should work decently well for some of the simpler cases - after all, the goal was never to replicate `Matplotlib` or `pyqtgraph` and the like. In my opinion, the biggest issue with *not* finishing this particular PR are the docs and deprecation notes.

How do you want to proceed? Some options are:

+ Merge as-is, ignoring all the proposed improvements in the PR note.
+ Close this PR and open a new one dedicated to only the docs and deprecation notes.
+ Close this PR and not merge anything.
+ Cherry-pick some commits
+ Other

RobinD42:
I'll leave it up to you.  If the current changes make sense to include we can merge this one, but it would also be nice to not lose track of the remaining todo items in case you or somebody else would like to work on them in the future. Maybe you could create an Issue with those items, links to this series of PRs and maybe a bit of explanation of what you had in mind for the remaining steps if they are not clear.

dougthor42:
Great idea. I've gone ahead and made separate issues for each item and updated the original PR message accordingly.

The new issues are:
+ #512
+ #513 
+ #514 
+ #515 
+ #516
+ #517 
+ #518
+ #519 

I've removed the WIP from the title and this is ready for review.

RobinD42:
Thanks.

aKummur:
Does this update allow a plot to have multiple scales like the one in - https://matplotlib.org/gallery/api/two_scales.html ?

dougthor42:
TBH I don't remember, haha. My gut tells me no, but let me investigate a bit.

aKummur:
@dougthor42 Thank you for confirming.

PR #119, opened 2016-07-20
Allow either cairocffi or PyCairo to be used from Phoenix

RobinD42:
OSX and Windows are working.  GTK coming next.


RobinD42:
https://github.com/SimonSapin/cairocffi/pull/86 is needed for OSX support.

PR #125, opened 2016-08-02
Add wrapper and unittests for wxPalette

swt2c:


RobinD42:
It would be better if the wxPalette wrappers could be more pythonic.

* `GetPixel` should allow integers, not just values from a bytearray.  If you tweak the parameter types to be "byte" then that should take care of it, as there is already a typedef for that which sets the SIP annotation `pyInt`. 

* `GetRGB` should should return the RGB values as a tuple of integers, and raise an exception (`ValueError` or perhaps `IndexError`) if the C++ method returns `false`.  One way to do this would be to ignore the existing method and use `addCppMethod` to add a replacement with the same name.

* The 3rd constructor and the `Create` should be able to take any Python sequences of integer values < 256, and the `n` parameter does not need to be in the API at all as Python already knows the length.  For `Create` that can be handled like the above.  For the constructor that can be tricky, I'll give that some more thought.

swt2c:
OK.  I'll make an attempt at doing those.

BTW, do waf partial rebuilds work for you?  It seems like every time I do a partial etg, then sip, when I do a build_py it seems to want to rebuild everything.

RobinD42:
There is automatic dependency checking, so if some changes causes one of the .h files to be modified then all source modules that #include that file will be rebuilt.  Unfortunately all item names (names of classes, functions, parameters, etc.) are defined in the .h file so any change, addition, reordering, etc. of the names will cause a rebuild of everything in that extension module.

swt2c:
Yeah, after I wrote that, I tried again and it did a quick rebuild, so I guess it might have been one of the headers.

Did you have any more thoughts on what to do with the constructor?  I thought about trying to call the re-implemented addCppMethod Create() but it seems like that may not be possible.

To be honest, this seems like it is starting to turn into a lot of effort for a little used class.

RobinD42:
Go ahead and make the other changes and I'll do something about the constructor before merging the PR.

swt2c:
Updated.

RobinD42:
Thanks.  The trouble I thought I remembered about replacing the constructor turns out to not apply in this case (no pure virtuals in this class).  So I've implemented it, refactored a few other things, and will merge later today.

PR #40, opened 2016-03-01
pdfviewer (old pr109)

RobinD42:
Moved from https://github.com/RobinD42/Phoenix/pull/109

PR #127, opened 2016-08-04
4 August 2016 Phoenix updates for wx.lib.pdfviewer and demo/PDFViewer

david-hughes:
demo/viewer.py and viewer_basics.py removed - was demo of obsolete wx.lib.vtk

wx.lib.pdfviewer updated for Phoenix and Python 3. Now requires PyMuPDF or PyPDF2. The latter renders using Cairo if present, otherwise wx.GrahicsContext.

Tested with 32 bit Python 2.7 and 64 bit Python 3.4 under Windows 10 and Python 2.7 & 3.5 under OS X
All combinations of imported package options tested - except for Cairo under OS X :(

This pull request incorporates all the changes in request  #40 (old PR109) made by Werner except for two moves to unittests of some unit_tests that were embedded in 3rd party code in wx.lib.pdfviewer. It is intended to produce a comprehensive set of tests for this package using pytest


wx/lib/pdfviewer/bezier.py
```diff
@@ -58,7 +58,7 @@ def compute_points(controlpoints, nsteps=30):
     """
     Input 4 control points as :class:`RealPoint` and convert to vec2d instances.
     compute the nsteps points on the resulting curve and return them
-    as a list of :class:`wx.Point`
+    as a list of :class:`Point`
```
RobinD42: The intra-doc links like this should keep the fully qualified names, otherwise there can be ambiguous names that Sphinx can't figure out and it just makes a (usually incorrect) guess.  I made a pass over everything a while back fixing those items that Sphinx gave warnings about, so maybe you just inadvertently overwrote my changes by starting with an older version of the files.  

While you are fixing these then it would be good to also add the leading `wx.` to the other links, for consistency.

PR #159, opened 2016-08-24
Pdfviewer2 - update buttonpanel.py and viewer.py

david-hughes:
Only the last commit is relevant. The others were created by me bumbling around trying to synchronise my local fork with wxWidgets/Phoenix master

wx/lib/pdfviewer/buttonpanel.py - Disable all controls until a PDF file has been opened

 wx/lib/pdfviewer/viewer.py - Scale the font metrics when using Cairo as well a scaling both the metrics and the font size when using wx.GraphicsContext (MSW only)


RobinD42:
I cherry-picked the 4th commit into another branch and merged that instead.  The change to the wxWidgets submodule revision would have been incorrect.

PR #166, opened 2016-09-06
dcgraphics.py, bezier.py and vec2d.py no longer needed for rendering …

david-hughes:
…to printer DC

wx.GraphicsContext, Cairo.GraphicsContext and MuPDF page-Pixmap can now be
created directly in a printer DC in pdfviewer.py (unlike originally in 2009).
Cacheing of page bitmaps now removed - jerky scrolling no longer observed.
SetZoom modified to stay on the same page after scale changed.

These changes originated in a user's report of fonts being displayed / printed at the wrong size
when using PyPDF2 under Windows. Although I have been unable to resolve that problem - which I will raise as a separate issue - it has resulted in tidying up and simplifying the code for output to a printer DC. I have forgotten why it was necessary to do it the way it was.


RobinD42:
There is a module in unittests that is testing vec2d that should also be removed.

david-hughes:
It would also be possible to remove the skipping of the whole pdfviewer unittest under OS X if the line 
1070     boxed = '--boxed' if not isWindows else ''
in build .py was amended. The tests work otherwise.

[Build-error-msg-osx.txt](https://github.com/wxWidgets/Phoenix/files/461436/Build-error-msg-osx.txt)

RobinD42:
Do you have the `pytest-xdist` package installed?  It is the one that provides the `--boxed` cli arg.

swt2c:
I also have problems when running the tests from build.py because of the --boxed.  When I run then manually with py.test --boxed, it works.  I've been meaning to look into that further.

david-hughes:
I have installed all the requirements on the Mac as well and it works fine with --boxed now. I have left the pdfviewer unittests all skipped under OS X as Python crashes (segmentation fault 11) during or after test_lib_pdfviewer_loadFile running both from build.py and pytest.py directly within WingIDE and I haven't been able to pinpoint the cause.

swt2c:
The pdfviewer tests crash on Linux too...

swt2c:
The thing with `python3 -m pytest --boxed` not working seems to be only a python3 issue, as python2 works.  I reported it as a pytest bug.

PR #180, opened 2016-10-04
Update Image Demo

Metallicow:
Update image demo images and show most all methods.

Screenie
image_screenie


RobinD42:
There are some good ideas here, but I think there are some issues that need to be addressed.  

First is the layout.  It seems a little confusing and too dense to me. I expect that newbies especially will not have any idea what is being demonstrated and how to get what is being demoed into their own applications.  Secondly, static widgets on OSX do not display tooltips, so all of the extended info you are trying to show will not be seen even if the user knows to try to show it by hovering over the images. And, IMO, even on the other platforms it's not a good idea to hide that in tooltips because it is non-intuitive.

So here is my suggestion: in a vertical box sizer add the following:

1. A "wx.Image" static text, with a larger font, to serve as a title

1. Perhaps some (brief) explanatory text describing what is on the screen

1. a "Bitmap Types" static text with a medium sized font to serve as a section title

1. A two column flex grid sizer with a row for each supported bitmap type, where the first column is a static text with info like you are currently putting in the tooltip for the bitmap types, and the second column is the static bitmap showing the image.

1. Another section title "Image Modification Operations"

1. Another 2 column flex grid sizer with the first row containing the color and file pickers in the first column, (using a nested sizer so they are both in the same cell) and the display of the unmodified image in the 2nd column.  Then the remaining rows would be the text describing the modification that you currently have in the tooltips and the modified image.

Does that make sense to you?

RobinD42:
Also, the duplicated code in `__init__` and `ChangeModdedImages` should be refactored into a separate method that is called from both locations.

Metallicow:
Sorry, I didn't realize the tooltips don't show on OSX. 
Yea, that would sorta explain it a bit better I think. I'll play around with your layout descriptions and see how it turns out.

Metallicow:
OK, this is a bit more better layed out now.

PR #187, opened 2016-10-18
Fix UltimateListCtrl FONTWEIGHT

Lokla:
NORMAL in Font Call should have been FONTWEIGHT, not FONTSYLE preventing UltimatListCtrl from working


Metallicow:
Correct. Sorry my typo.

PR #193, opened 2016-11-07
pycolourchooser now, er, works

tekhedd:
Submitted for your approval, I have apparently finished implementing pycolourchooser. I assume that some of its code was removed over time to get it to compile, as it did not appear to be completely implemented.

It still does not want to resize, but otherwise works. 

Tested on osx. I'll definitely have to test on Windows this week, but in any case...

```
import wx.lib.colourchooser.pycolourchooser as PCC
PCC.main()
```

The other native-wrapping colourselect controls were giving me fits, so here you go. Share and enjoy.

### DIFF
--- a/wx/lib/colourchooser/canvas.py
+++ b/wx/lib/colourchooser/canvas.py
@@ -80,13 +80,16 @@ class Canvas(wx.Window):
     """
     def __init__(self, parent, id,
                  pos=wx.DefaultPosition,
-                 size=wx.DefaultSize,
-                 style=wx.SIMPLE_BORDER):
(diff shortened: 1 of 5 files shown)


tekhedd:
More testing showed that click-dragging on the palette required CaptureMouse() to work correctly on windows. The V slider palette window is now clickable, although not nearly as fancy (it should probably also do a CaptureMouse().).  Both the palette and slider window now correctly size to fit their content. 

an event is now sent in real time as the color changes, to make previews more pleasurable.

This is feeling more and more like a stone soup experiment. ;)

PR #210, opened 2017-01-22
WIP: mustHaveApp

RobinD42:
Adds the ability to mark classes, methods, or functions to require that the wx.App object already be created, and to raise an exception if it hasn't been.  

There is still some work to be done here, some temporary workarounds to be undone, and for the new changes in SIP to stabilize.


RobinD42:
Oops, included some wrong commits, will resubmit...

Issue #206, opened 2017-01-15
GStreamer 0.10 is not available in Ubuntu 16.10


ccfiel:
I got this error doing a pip install wx using ubuntu 16.10. I did also try to install 

```
wget http://fr.archive.ubuntu.com/ubuntu/pool/main/g/gst-plugins-base0.10/libgstreamer-plugins-base0.10-0_0.10.36-1_amd64.deb
wget http://fr.archive.ubuntu.com/ubuntu/pool/universe/g/gstreamer0.10/libgstreamer0.10-0_0.10.36-1.5ubuntu1_amd64.deb
sudo dpkg -i libgstreamer*.deb
```

Still got the same problem. Any ideas?

```
  chhecking for cairo_push_group... yes
  checking for GST... configure: WARNING: GStreamer 0.10 not available, falling back to 0.8
  checking for GST... configure: WARNING: GStreamer 0.8/0.10 not available.
  configure: error: GStreamer not available
  Error running configure
  ERROR: failed building wxWidgets
  Traceback (most recent call last):
    File "build.py", line 1184, in cmd_build_wx
```

eldarkg:
@ccfiel You need to install next packages from [Ubuntu 16.04](http://packages.ubuntu.com/xenial/):
gir1.2-gst-plugins-base-0.10_0.10.36-2_amd64.deb
gir1.2-gstreamer-0.10_0.10.36-1.5ubuntu1_amd64.deb
libgstreamer0.10-0_0.10.36-1.5ubuntu1_amd64.deb
libgstreamer0.10-dev_0.10.36-1.5ubuntu1_amd64.deb
libgstreamer-plugins-base0.10-0_0.10.36-2_amd64.deb
libgstreamer-plugins-base0.10-dev_0.10.36-2_amd64.deb

swt2c:
wxWidgets 3.0 should now have gstreamer 1.0 support via 501b377.

@RobinD42 can you please bump the wxWidgets subproject commit to pick up these changes?

RobinD42:
@swt2c, yep. I did it as soon as I noticed your change had been merged.  I'm waiting on one more build and a test of an additional change and then I'll push them.

PR #212, opened 2017-01-22
Created options for build.py for ensuring presence of required packages

mesalu:
# Overview
## Description

Adds the modes `checkdeps` and `instdeps` to build.py.
`checkdeps` can be used to list any missing packages, exiting the build should necessary packages be missing.
`instdeps` will attempt to install missing packages using the distributions native package manager. 

Internally packages are separated into two main groups, 'required' and `recommended`.
`recommended` packages are only libraries which wx can substitute for itself, such as libpng, libtiff, etc.
`required` packages are packages that will break Phoenix's build if not present such as glu, gtk2/3, webkitgtk.  If a required package is missing when checkdeps is used then the build will exit, if instdeps is used then it will prevent moving forward until it is installed.

## Usage
The common case I tested was for a new fresh-repository build:
`python build.py checkdeps dox etg --nodoc sip build`
If any required packages are missing the build will exit as expected.

`python build.py instdeps <other args>`
Will attempt to invoke the package manager as non-root and exit with the package manager's "you're not root" message

`sudo python build.py instdeps <other args>`
will ensure the user is aware they are running the script as root, and then proceed to use the native package manager to install missing packages.

## Side comments
I didn't test moving the new switches around within the command, I always used them first. I think it should be fine so long as they preceed `build`.

should the new modes need to exit they'll directly invoke sys.exit, I didn't read the rest of build.py enough to know if this direct exit will mess with any running processes or if it will affect user experience

I didn't include python-dev packages as requirements, which usually isn't a problem. I think it only affected me on a cent6 test, though I don't remember for sure. I think its fair to assume that the necessary python development files are already present for the common user. 

apt-get will internally adjust the name of some libraries, such as libpng and libtiff to be versioned. This will cause these libraries to show up as missing regardless of their versioned package being present.

I didn't spend too much time investigating, but I think yum doesn't exit with a non-zero value should an install fail. 
More on CentOS: I didn't include the installation of the user respository as a requirement for cent, this will cause webkitgtk and webkitgtk-devel to be missing but uninstallable. Making a comment about it somewhere cent users will see might prove beneficial

While testing this feature I realized that Phoenix does not work for Ubuntu16.10, as gstreamer 0.10 is not available causing wx's configuration to fail. The package may be available on a user repository somewhere, though I'm not sure. 
Edit: looks like issue #206 covers this

# The Manifest
A yaml manifest file was created for translating necessary packages to distribution specific names. Each `required` package common to every distribution has its own entry in the manifest per distribution.  An additional field "other" is provided for listing packages specific to a specific distribution.
The manifest is organized by first distribution, then version, then package fields, which will be read as strings and split by the python code. this leaves the manifest very human readable, so hopefully its easy to pick up and maintain.

### DIFF
--- a/build.py
+++ b/build.py
@@ -134,6 +134,9 @@ Usage: ./build.py [command(s)] [options]
       test          Run the unit test suite
       test_*        Run just the one named test module
 
+      checkdeps     Linux only. Print missing necessary packages
+      instdeps      Linux only. Attempt to install missing necessary packages
+
       clean_wx      Clean the wx parts of the build
       clean_py      Clean the wxPython parts of the build
       clean_sphinx  Clean the sphinx files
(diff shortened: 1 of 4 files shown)


RobinD42:

mesalu:
Closing for remodeling.
Will return later with a less obtrusive, more useful, and cleaner alternative.

Issue #207, opened 2017-01-16
Fix PyEvent.Clone


RobinD42:
The python `copy` module can't be used on wrapped classes, but it should not be needed for `PyEvent` anyway.  See http://stackoverflow.com/questions/41642788/error-occurred-when-hovering-on-a-wxpython-tab

PR #215, opened 2017-01-25
enable pickling of PyEvent objects

mesalu:
Adds __setstate__ and __getstate__ methods to wx.PyCommandEvent and wx.PyEvent objects, enabling them to be pickled.
This allows copying of object instances, fixing the issue detailed in #207 
Also, this preserves that python magic ;-)

### DIFF
--- a/etg/pyevent.py
+++ b/etg/pyevent.py
@@ -82,13 +82,13 @@ def run():
             """)
 
     cls.addPyMethod("__setstate__", "(self, state)",
-        doc = """Sets internal state of PyEvent object to allow pickling""",
+        doc = """Sets internal state of PyEvent object. This allows PyEvent to be reconstructed as part of the pickling process""",
         body = """\
             self.__dict__.update(state)
             """)
 
     cls.addPyMethod("__getstate__", "(self)",
-        doc = "Returns internal state. Precense enables pickling",
+        doc = "Returns this object's internal state. This is used as part of the pickling process",
         body = """\
             return self.__dict__
             """)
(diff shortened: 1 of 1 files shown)


etg/pyevent.py
```diff
@@ -81,6 +81,18 @@ def run():
...
             """)
 
+    cls.addPyMethod("__setstate__", "(self, state)",
+        doc = """Sets internal state of PyEvent object to allow pickling""",
+        body = """\
+            self.__dict__.update(state)
+            """)
+
+    cls.addPyMethod("__getstate__", "(self)",
+        doc = "Returns internal state. Precense enables pickling",
```
RobinD42: Typo: presence

etg/pyevent.py
```diff
@@ -138,6 +150,18 @@ def run():
...
```
RobinD42: Typo: presence

RobinD42: Typo: presence

RobinD42:
Looks good other than the typos. Thanks.

etg/pyevent.py
mesalu: Oops. 

Also, I'm still not 100% confident with etg, or whats going on cpp side / py side. 

The get/set state methods allow pickle, well, pickle the object. Because otherwise it can't be sure of the full-state of objects that overwrite __getattr__

Should the state control methods utilize the cpp side state using _getAtrDict()?

mesalu:
Also, I'm still not 100% confident with etg, or whats going on cpp side / py side. 

The get/set state methods allow pickle, well, pickle the object. Because otherwise it can't be sure of the full-state of objects that overwrite __getattr__

Should the state control methods utilize the cpp side state using _getAtrDict()?

mesalu:
In answering my own question:
Using _getAttrDict() in the `__setstate__` will cause a RuntimeError, as the reconstructed PyEvent object hasn't been fully instantiated yet, this happens a couple lines later in `Clone` anyways. Furthermore, any data not preserved by sending the object through `copy` will be restored by the aforementioned instantiation.

RobinD42:
Fixes #207

PR #216, opened 2017-01-28
PyEvent Clone fix, 2nd try

RobinD42:
Instead of using copy.copy to implement Clone, create the new instance with __new__

@mesalu, thoughts?


mesalu:
👍

PR #236, opened 2017-02-14
Ensure Tree Mixin text visibility against a white background in demo

mesalu:
Addresses demo side aspect of #201


mesalu:
As long as the native text color really is supposed to be white I'm not sure there's much for it other than to set text color to something visible against a white background whenever the TreeMixin is used.

PR #242, opened 2017-02-27
Use AppVeyor in addition to TravisCI to get merge validation for Windows

mesalu:
Allows appveyor to run builds on project. This performs the same role as TravisCI, except it covers windows rather than linux.
Currently configured to do four builds using 64 and 32 bit versions of python2.7 and python 3.6
No testing is currently being run, partly due to lack of boxed tests and partly to save build time.

Necessary appveyor config:
Need to set appveyor to use `.appveyor.yml` rather than `appveyor.yml`. This can be done in the general settings tab.
I suggest activating `Rolling Builds` in the same tab, to prevent a huge backlog of builds.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,29 @@
+# Samuel Dunn
+# Simple Appveyor configuration file for Phoenix
+version: "{build}"
+configuration: Release
+platform:
+- x64
+environment:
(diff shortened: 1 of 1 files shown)


mesalu:
I want to refine this a bit more.

PR #258, opened 2017-03-24
Removed unused variables

radarhere:


build.py
```diff
@@ -1686,7 +1685,6 @@ def cmd_sdist(options, args):
             copyFile(name, destdir)
 
     # Also add the waf executable
-    waf = getWafCmd()
```
RobinD42: This line is needed because it is possible that the waf executable has not been fetched yet when doing a sdist command.  We don't need to save the return value of the function, but it does need to be called.

radarhere: Okay, thanks, I've updated the commit.

Metallicow: Please place a line comment on that so this is known in code.

PR #268, opened 2017-04-11
etg: add eq+hash to value types

adrientetar:
I wanted to put wx.Colour in a dict(), but couldn't due to the lack of hash function.

There are probably more types that need \_\_hash__ defined (value types primarily) but this is a start.

r? @RobinD42

(By the way, instead of redefining a number of methods to use the internal tuple, I thought maybe wx.Colour could subclass tuple directly... I don't know if that would work, but is worth trying I think.)

### DIFF
--- a/etg/colour.py
+++ b/etg/colour.py
@@ -165,6 +165,8 @@ def run():
 
 
     # Add sequence protocol methods and other goodies
+    c.addPyMethod('__eq__', '(self, other)',       'return isinstance(other, type(self)) and self.Get() == other.Get()')
+    c.addPyMethod('__hash__', '(self)',            'return hash(self.Get())')
     c.addPyMethod('__str__', '(self)',             'return str(self.Get())')
     c.addPyMethod('__repr__', '(self)',            'return "wx.Colour"+str(self.Get())')
     c.addPyMethod('__len__', '(self)',             'return len(self.Get())')
--- a/etg/gbsizer.py
+++ b/etg/gbsizer.py
@@ -52,6 +52,8 @@ def run():
         briefDoc="Set both the row and column properties.")
 
     # Add sequence protocol methods and other goodies
+    c.addPyMethod('__eq__', '(self, other)',       'return isinstance(other, type(self)) and self.Get() == other.Get()')
+    c.addPyMethod('__hash__', '(self)',            'return hash(self.Get())')
     c.addPyMethod('__str__', '(self)',             'return str(self.Get())')
     c.addPyMethod('__repr__', '(self)',            'return "wx.GBPosition"+str(self.Get())')
     c.addPyMethod('__len__', '(self)',             'return len(self.Get())')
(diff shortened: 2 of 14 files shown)


RobinD42:
> There are probably more types that need __hash__ defined (value types primarily) but this is a start.

Yes, this is good.  Please also add it for `wx.Point`, `wx.Size`, `wx.Rect` and anything else you see that follows this pattern. Adding some simple tests to the appropriate unittest modules would also be a good idea.

> (By the way, instead of redefining a number of methods to use the internal tuple, I thought maybe wx.Colour could subclass tuple directly... I don't know if that would work, but is worth trying I think.)

Interesting, but I don't think it's possible since the types are implemented in C++ by SIP. (The code you are modifying is just monkey-patched on to those extension types.)

adrientetar:
Okay, I added \_\_eq__ and \_\_hash__ to all value types and I added corresponding unit tests. Let's see if the CI likes it.

(Had to define \_\_eq__ as well because the default implementation would match the tuple with the custom class, which we don't want in order to avoid duplicates entries in hash-based data structures).

RobinD42:
I'm not sure I follow your explanation of the need for `__eq__`. Is it just that wx.Point, wx.Size and such are basically interchangeable via sequence conversions? If so, that is fixable in a better way (I went ahead and did it in #272, I've been meaning to do it for a while.) If not, then please explain further. I'd really like to keep the ability to compare these types with tuples or lists.

adrientetar:
Well basically nothing tells you that there wouldn't be hash collisions between objects of different types, since each object type has its own way of calculating the hash. For this reason, hash-based data structure discriminate between object types using eq _then_ they use the hash itself.  
Therefore the eq function must return false for objects that are of a different class.

There would, besides, be a collision between e.g. wx.Colour(0, 0, 0, 0) and (0, 0, 0, 0) when putting these objects in a dict, set etc. putting eq aside.

This posts puts it down in more details: http://stackoverflow.com/a/19073010

> I'd really like to keep the ability to compare these types with tuples or lists.

Why though? Just calling Get() to obtain the tuple is short more correct, if you want to run an equality test. Otherwise you could just compare color to color.

RobinD42:
Okay, thanks. I wasn't aware of that aspect of how the hashing works for keys and sets.  Makes sense.  BTW, it looks like `NotImplemented` should be returned if the type isn't compatible. Also, some of these types already have code for adding implementations of `__eq__` and `__ne__` wrapping the C++ operators that should be removed.

>> I'd really like to keep the ability to compare these types with tuples or lists.
> Why though?

Just the old "because that's the way it's always been" excuse. ;-)

RobinD42:
One more thought.  These objects are mutable, which IIUC, will cause problems if the values change after the object has been used as a key in a dictionary, correct?

adrientetar:
Ah yes, though these types have tuples going on they are mutable... I've been using PyQt5 and just checked, the value types aren't hashable there.

For the record, my use-case is when generating certain bitmaps etc. I have a cache (with a dict) parametrized on attributes like color. I guess I'll just take the tuple out of the color object when calling it then make a new color object with the tuple inside the factory.

RobinD42:
See the TODO added in 254a9b6 for another idea for this that would be a little better than just using the `obj.Get()` tuples in the dictionaries.

adrientetar:
Another idea would be to create a new python object each type it's mutated, but for now I can work-around this easily.

dougthor42:
Additional reference:
[wxPython-dev: wx.Colour - not hashable in Phoenix/Py3](https://groups.google.com/d/topic/wxpython-dev/NLd4CZv9rII/discussion)

Issue #243, opened 2017-03-03
wxp.3.0.3 AUI panes do not resize


psionprime:
I thought I had just messed up a style bit but the sample from:

https://wxpython.org/Phoenix/docs/html/wx.lib.agw.aui.html

does not allow the panes to move either.

I'm happy to dig into it on my own if someone can point me in the right direction please.

I've been developing many apps with the library with great success but apparently never tried to move the panes before now.

Thanks for your efforts !

stevenj:
I had this problem as well.  Look at the code to AUI_DockingWindowMgr.py in the demo suite.  Its resizing works. (Although it gives no feedback until you stop dragging.)  The difference is:

import wx.aui as aui
instead of:
import wx.lib.agw.aui as aui

try that.

Also, if you want resizing to give some active feedback try this:
self._mgr.SetFlags(self._mgr.GetFlags() ^ aui.AUI_MGR_LIVE_RESIZE)

acollange:
Are you using GTK ?
Because I have found something strange in the Update() method of AuiManager (agw version).

`    def Update(self):
        if '__WXGTK__' in wx.PlatformInfo:
            self.Bind(wx.EVT_WINDOW_CREATE, self.DoUpdateEvt)
        else:
            self.DoUpdate()

    def DoUpdateEvt(self, evt):
        wx.CallAfter(self.DoUpdate)`

The GTK code branch is useless.
Calling DoUpdate() instead of Update() should fix your problem.

Is somebody able to tell us what is the purpose of this Bind event here ?

psionprime:
I am using GTK (Linux Mint 18.1) I think this should remain open as the tips indicate a work-around. I am curious to the thinking of the maintainers if the tips actually are the proper method and the example needs updated or it is something else.... just the nomenclature awg vs aui and Update vs DoUpdate is ambiguous.

Text and my ability to render it sucks so please don't take what I pointed out as negative. I, for one, very much appreciate the Phoenix team and others' efforts.

Thanks for the tips !

RobinD42:
> Is somebody able to tell us what is the purpose of this Bind event here ?

[...]

PR #270, opened 2017-04-12
Update framemanager.py

acollange:
After digging into Aui update issue under GTK (#243), I propose to move the bind on `EVT_WINDOW_CREATE` into the `AuiManager` constructor and call `wx.CallAfter(self.DoUpdate)` in the `Update` method.

Like @RobinD42 said, it is mandatory to wait for `EVT_WINDOW_CREATE` under GTK, and my own experience on my project shown it is true !
Using `wx.CallAfter` to call `self.DoUpdate` seems also mandatory to avoid the debug message below:

`Debug: ScreenToClient cannot work when toplevel window is not shown`

### DIFF
--- a/wx/lib/agw/aui/framemanager.py
+++ b/wx/lib/agw/aui/framemanager.py
@@ -4218,6 +4218,7 @@ class AuiManager(wx.EvtHandler):
         self.Bind(wx.EVT_TIMER, self.OnHintFadeTimer, self._hint_fadetimer)
         self.Bind(wx.EVT_TIMER, self.SlideIn, self._preview_timer)
         self.Bind(wx.EVT_WINDOW_DESTROY, self.OnDestroy)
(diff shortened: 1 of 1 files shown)


Nutznieser:
Please see the issue #317. There are some unfavorable side effects with this code change.

I guess this code change was only important for using under GTK? My suggestion would be the following modification:
```
if '__WXGTK__' in wx.PlatformInfo:
    self.Bind(wx.EVT_WINDOW_CREATE, self.DoUpdateEvt)
```
And please, anybody have to check if the issue #317 also leads to the same problems under GTK. If so, then here is still more work to do.

acollange:
Sorry, that's my fault... I've forgotten the `if '__WXGTK__' in wx.PlatformInfo` condition in my patch.

> And please, anybody have to check if the issue #317 also leads to the same problems under GTK. If so, then here is still more work to do.

I was not able to check it yet, but it is highly probable with the new and the previous code.

PR #271, opened 2017-04-12
Classic wx.YieldIfNeeded() conversion error in CustomTreeCtrl 

acollange:
'Classic' wx.YieldIfNeeded() was translated into wx.SafeYield(), but it is not equivalent.
Cause a 'recursive Yield call' error in one of my project which works fine with Classic.
Change it into wx.SafeYield(onlyIfNeeded=True) resolve the issue.

### DIFF
--- a/wx/lib/agw/customtreectrl.py
+++ b/wx/lib/agw/customtreectrl.py
@@ -5867,7 +5867,7 @@ class CustomTreeCtrl(wx.ScrolledWindow):
             if wx.Platform in ["__WXMSW__", "__WXMAC__"]:
                 self.Update()
         else:
-            wx.SafeYield()
+            wx.SafeYield(onlyIfNeeded=True)
 
         # now scroll to the item
         item_y = item.GetY()


RobinD42:
Actually, the equivalent of `wx.YieldIfNeeded` would be `wx.GetApp().Yield(True)`. Would that be a better fit here?

BTW, I hadn't realized that `wx.YieldIfNeeded` had gotten lost in Phoenix. I'll add it back in. Thanks.

acollange:
Yes `wx.GetApp().Yield(True)` would fit perfectly here. The important thing for me is the `onlyIfNeeded` thing, not the `Safe` one.
But if you plan to reintroduce `wx.YieldIfNeeded`, then just throw away this request and restore the original 'Classic' version. `wx.YieldIfNeeded` was used multiple times in customtreectrl.py (and probably elsewhere).

RobinD42:
`wx.YieldIfNeeded` should be back in tonight's snapshot build, please make the switch, test and update this PR. Please don't just copy the Classic versions of the modules as there will be other changes that will need to be kept.  Thanks.

RobinD42:
Switched to using `wx.YieldIfNeeded` in  customtreectrl.py in 5a669d54622d46635ba0b3619bbc9877809fa5b6

PR #277, opened 2017-04-15
Add a GetIM() method to Colour, Point, Size, Rect and other ‘value’ types

RobinD42:
Add a GetIM() method to Colour, Point, Size, Rect and other ‘value’ types.

GetIM() returns a replicant of the original object that is immutable, so it can be used as a dictionary key, or etc., but still gives access to the properties by name, which using the Get() method to get a tuple of the values does not do.

See #268

### DIFF
--- a/etg/colour.py
+++ b/etg/colour.py
@@ -163,6 +163,8 @@ def run():
         briefDoc="""\
         Returns the RGB intensity values as a tuple, optionally the alpha value as well.""")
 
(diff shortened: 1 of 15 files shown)


dougthor42:
Additional reference:
[wxPython-dev: wx.Colour - not hashable in Phoenix/Py3](https://groups.google.com/d/topic/wxpython-dev/NLd4CZv9rII/discussion)

PR #313, opened 2017-04-20
Flag icon updates

Metallicow:
Optimized PNGs
README.rst with table
PNG Extraction code


resources/extracted_flagart_bitmaps/README.rst
```diff
@@ -0,0 +1,322 @@
...
+
+
+Sanity checks
+-------------
+ * https://en.wikipedia.org
+ * http://flagpedia.net/
+
+
++--------------------------+------------+-------------------------------------------+--------------------------------------------------------------------------------------------------+
+| IMG                      | PNG        | Country Name                              | Flag Reference                                                                                   |
```
RobinD42: Please reduce the width of this table to the table readable as text in a window that is 100 characters wide, or at least as close to 100 as you can get it.

RobinD42:

Metallicow:
@RobinD42  It doesn't look like I can edit labels. maybe a repo setting... I'll see if I can finish this tonight...

Metallicow:
This is a very small but helluva feat to accomplish. It probably would be a bit a lot different nowadays..... It may be better for other individuals to conquer with smaller portions. Here is a sample of that works at pulltime. Thanks.

Do it once and do it right 1000 flags, then not much problems singlefold.
Do it the next day after and its like staying 3 nights up with no sleep.
Good luck.

PR #323, opened 2017-04-24
fixes for DVC_CustomRenderer Demo

jensgoe:
- GetValueFromEditorCtrl must return tuple
- SetValueByRow must return a boolean value


jensgoe:
I am not sure if you should merge these changes are adapt the interfaces of these two methods. GetValueFromEditorCtrl  MUST return a tuple now, that was quite confusing to mee. Especially the first element of the tuple has no effect. 
The return value of SetValueByRow  seems to have no effect two. 

Anyway theses changes fixes the demo for DVC_CustomRenderer.

RobinD42:
I've changed things in a different way in #329. Thanks for the reminder.

PR #375, opened 2017-05-30
Suppress `framemanager` warnings and errors

pauldmccarthy:
This PR addresses a number of issues when using the `framemanager.py`, suppressing harmless, but annoying, warnings and errors. The test program in the first comment can be used to demonstrate all of the warnings and errors that are addressed.

`ClientToScreen cannot work when toplevel window is not shown`
--------------------------------------------------------------

    ClientToScreen cannot work when toplevel window is not shown
    ScreenToClient cannot work when toplevel window is not shown

These warnings occur under GTK when floating, moving, and docking panes in an AUI managed window.  They are output whenever a call to `wxWindow.ClientToScreen` or `ScreenToClient` (direct or indirect) is made on a window which is not visible.  They seem to occur due to `AuiFloatingFrame` and `AuiDockingGuide` instances being used before they have actually appeared on screen.

The suggested fix suppresses _most_ of these warnings. However, it is not perfect, because the GTK implementation of `IsShownOnScreen` seems to lie on occasion (i.e. it returns `True` when it shouldn't).

`RuntimeError: wrapped C/C++ object of type AuiNotebook has been deleted`
-------------------------------------------------------------------------

      File "/home/paulmc/.virtualenvs/fsleyes3.5/lib/python3.5/site-packages/wx/core.py", line 3133, in <lambda>
        lambda event: event.callable(*event.args, **event.kw) )
      File "/home/paulmc/.virtualenvs/fsleyes3.5/lib/python3.5/site-packages/wx/lib/agw/aui/framemanager.py", line 6370, in DoUpdate
        self.GetManagedWindow().Freeze()
    RuntimeError: wrapped C/C++ object of type AuiNotebook has been deleted

This error occurs when a tabbed pane (a pane in an `AuiManager`, which has been grouped with one or more panes into a notebook) is either dragged out or closed.  Adding a guard at the beginning of `DoUpdate` suppresses this error.

`wx._core.wxAssertionError: C++ assertion "GetEventHandler() == this" failed`
-----------------------------------------------------------------------------

    wx._core.wxAssertionError: C++ assertion "GetEventHandler() == this" failed at /tmp/pip-build-jnz8ex_j/wxpython/ext/wxWidgets/src/common/wincmn.cpp(478) in ~wxWindowBase(): any pushed event handlers must have been removed

    During handling of the above exception, another exception occurred:

    SystemError: <class 'wx._core.EraseEvent'> returned a result with an error set

This is a trickier one. It occurs in two situations that I have found:

 - When a floating pane is programmatically detached from an `AuiManager`
 - When an Aui-mnaged top level frame is closed by the user.

It occurs because, in these situations, the `AuiManager.UnInit` method does not get called. This leaves the `AuiManager` instance on the event handler stack, causing the assertion error to occur. It seems that the `AuiManager.UnInit` method **must** be called when its managed window is destroyed to prevent this error. However, the guard in `AuiManager.OnDestroy` seems to prevent this from happening (Use the `OnDestroy` monkey patch in the test program to see this in action):

    if self._frame == event.GetEventObject():
        self.UnInit()

I'm not sure what is going on but, in my experimenting, the event object is *never* equal to the managed window, so `UnInit` never gets called. An inelegant but effective solution to this problem is to:

  1. Explicitly call `UnInit` when `DetachPane is called on a floating pane.

  2. Listen to the `EVT_CLOSE` event in addition to `EVT_WINDOW_DESTROY`.

This does the trick for me.

### DIFF
--- a/wx/lib/agw/aui/framemanager.py
+++ b/wx/lib/agw/aui/framemanager.py
@@ -4218,6 +4218,7 @@ class AuiManager(wx.EvtHandler):
         self.Bind(wx.EVT_TIMER, self.OnHintFadeTimer, self._hint_fadetimer)
         self.Bind(wx.EVT_TIMER, self.SlideIn, self._preview_timer)
         self.Bind(wx.EVT_WINDOW_DESTROY, self.OnDestroy)
+        self.Bind(wx.EVT_CLOSE, self.OnClose)
         if '__WXGTK__' in wx.PlatformInfo:
             self.Bind(wx.EVT_WINDOW_CREATE, self.DoUpdateEvt)
 
@@ -4598,7 +4599,20 @@ class AuiManager(wx.EvtHandler):
                     klass.RemoveEventHandler(handler)
 
 
+    def OnClose(self, ev):
+        """Called when the managed window is closed. Makes sure that :meth:`UnInit`
+        is called.
+        """
+
+        ev.Skip()
+        if ev.GetEventObject() == self._frame:
+            wx.CallAfter(self.UnInit)
+
+
     def OnDestroy(self, event) :
+        """Called when the managed window is destroyed. Makes sure that :meth:`UnInit`
+        is called.
+        """
 
         if self._frame == event.GetEventObject():
             self.UnInit()
@@ -4968,6 +4982,7 @@ class AuiManager(wx.EvtHandler):
                     # reparent to self._frame and destroy the pane
                     p.window.Reparent(self._frame)
                     p.frame.SetSizer(None)
+                    p.frame._mgr.UnInit()
                     p.frame.Destroy()
                     p.frame = None
 
--- a/wx/lib/agw/aui/framemanager.py
+++ b/wx/lib/agw/aui/framemanager.py
@@ -6360,6 +6360,10 @@ class AuiManager(wx.EvtHandler):
         must be called. This construction allows pane flicker to be avoided by updating
         the whole layout at one time.
         """
+
+        if not self.GetManagedWindow():
+            return
+
         if '__WXGTK__' in wx.PlatformInfo:
             self.GetManagedWindow().Freeze()
         self._hover_button = None
--- a/wx/lib/agw/aui/framemanager.py
+++ b/wx/lib/agw/aui/framemanager.py
@@ -3090,7 +3090,7 @@ class AuiFloatingFrame(wx.MiniFrame):
         :param `event`: a :class:`wx.SizeEvent` to be processed.
         """
 
-        if self._owner_mgr and self._send_size:
+        if self._owner_mgr and self._send_size and self.IsShownOnScreen():
             self._owner_mgr.OnFloatingPaneResized(self._pane_window, event.GetSize())
 
 
@@ -6869,7 +6869,8 @@ class AuiManager(wx.EvtHandler):
             if guide.dock_direction == AUI_DOCK_CENTER:
                 guide.host.ValidateNotebookDocking(paneInfo.IsNotebookDockable())
 
-            guide.host.UpdateDockGuide(mousePos)
+            if guide.host.IsShownOnScreen():
+                guide.host.UpdateDockGuide(mousePos)
 
         paneInfo.window.Lower()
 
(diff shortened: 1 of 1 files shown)


pauldmccarthy:
Test program:

```python
#!/usr/bin/env python

from collections import OrderedDict

import wx
import wx.lib.agw.aui as aui


import logging
logging.basicConfig()
# logging.getLogger('fsleyes_props.trace').setLevel(logging.DEBUG)



# Monkey-patch AuiManager to see what AuiManager.OnDestroy does.
def OnDestroy(self, event):

    print('')
    print('')
    print('[{}] OnDestroy'.format(id(self)))
    print('  frame:  {}'.format(self._frame))
    print('  evobj:  {}'.format(event.GetEventObject()))

    if self._frame == event.GetEventObject():
        print('  UnInit')
        self.UnInit()

    print('')
    print('')


# aui.AuiManager.OnDestroy = OnDestroy


class MyFrame(wx.Frame):

    def __init__(self):

        wx.Frame.__init__(self, None)

        # Test with default, aero, and whidbey docking guides

        # self.auiMgr = aui.AuiManager(self,
        #                              agwFlags=aui.AUI_MGR_ALLOW_FLOATING)
        self.auiMgr = aui.AuiManager(self,
                                     agwFlags=(aui.AUI_MGR_ALLOW_FLOATING |
                                               aui.AUI_MGR_AERO_DOCKING_GUIDES))
        # self.auiMgr = aui.AuiManager(self,
        #                              agwFlags=(aui.AUI_MGR_ALLOW_FLOATING |
        #                                        aui.AUI_MGR_WHIDBEY_DOCKING_GUIDES))

        centre   = wx.Panel(self)
        sizer    = wx.BoxSizer(wx.VERTICAL)
        addbtn   = wx.Button(centre, label='Add a pane')
        closebtn = wx.Button(centre, label='Close last pane')

        self.paneNum = 0
        self.panes   = OrderedDict()

        sizer.Add(addbtn,   flag=wx.EXPAND, proportion=1)
        sizer.Add(closebtn, flag=wx.EXPAND, proportion=1)

        centre.SetSizer(sizer)
        addbtn  .Bind(wx.EVT_BUTTON, self.onaddbtn)
        closebtn.Bind(wx.EVT_BUTTON, self.onclosebtn)

        paneinfo = aui.AuiPaneInfo().CentrePane()

        self.auiMgr.Bind(aui.EVT_AUI_PANE_CLOSE, self.onpaneclose)

        self.auiMgr.AddPane(centre, paneinfo)
        self.auiMgr.Update()


    def onaddbtn(self, ev):

        print('onaddbtn')

        pane     = wx.Panel(self)
        name     = 'Pane {}'.format(self.paneNum)
        text     = wx.StaticText(pane, label=name)
        sizer    = wx.BoxSizer(wx.VERTICAL)
        paneinfo = (aui.AuiPaneInfo()
                    .Name(name)
                    .Caption(name)
                    .Float()
                    .MinSize((100, 50))
                    .Resizable())

        sizer.Add(text, flag=wx.EXPAND, proportion=1)
        pane.SetSizer(sizer)

        self.auiMgr.AddPane(pane, paneinfo)
        self.auiMgr.Update()

        self.panes[name] = paneinfo
        self.paneNum    += 1


    def onclosebtn(self, ev):

        print('onclosebtn')

        if len(self.panes) == 0:
            return

        pane = self.panes.popitem()[1].window

        self.auiMgr.DetachPane(pane)
        wx.CallAfter(pane.Destroy)


    def onpaneclose(self, ev):

        print('onpaneclose')

        ev.Skip()

        paneinfo = ev.GetPane()
        pane     = paneinfo.window

        # EVT_AUI_PANE_CLOSE gets called twice
        # when a pane is closed by the user.
        # Not bothering to fix this at the moment.
        if self.panes.pop(paneinfo.name, None) is None:
            return

        self.auiMgr.DetachPane(pane)
        wx.CallAfter(pane.Destroy)
        wx.CallAfter(self.auiMgr.Update)


if __name__ == '__main__':

    app   = wx.App()
    frame = MyFrame()

    frame.Show()
    app.MainLoop()
```

PR #407, opened 2017-06-21
Fixing missing Return Value

astral451:
I tested this change locally and found that because it didn't have a return of the int it was swallowing all events in this spot.


RobinD42:
Thanks

PR #442, opened 2017-07-24
Add docs to the release tarball

ghisvail:
The PyPI tarball is the canonical source for building packages on Linux distributions such as Debian. It would be desirable to have the docs shipped in the sdist, so that a corresponding documentation package can be built on the Debian side.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -8,6 +8,7 @@ recursive-include   wx  **
 recursive-include   license  *.txt
 include             LICENSE.txt
 graft               docs
+prune               docs/sphinx/build
 
 recursive-exclude   wx  .git
 recursive-exclude   wx  *.pyc
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -7,6 +7,7 @@
 recursive-include   wx  **
 recursive-include   license  *.txt
 include             LICENSE.txt
+graft               docs
 
 recursive-exclude   wx  .git
 recursive-exclude   wx  *.pyc


RobinD42:
Unfortunately it's not going to be that easy. 

Currently the source tarball is built on the linux build machine so the line endings will be "proper" for most of the users of the source tarball. Meanwhile the docs are built on a Windows machine because there are some classes that are only available on Windows. If the doc builder was moved to Linux then when Sphinx is extracting docs it will skip those Windows-only classes due to import errors or other exceptions.

ghisvail:
> If the doc builder was moved to Linux then when Sphinx is extracting docs it will skip those Windows-only classes due to import errors or other exceptions.

Would that be an issue for Linux users? The missing documentation would be for classes they don't have access to, right?

RobinD42:
Correct, but the problem then becomes that we have two sets of documentation files in the release that are not identical, which could cause confusion. 

So here's another idea: I could include just the generated documentation source files in the source tarball, but leave out the final `.html` files.  Then when you (and other linux packagers) are building your packages you can add an extra step to run Sphinx on the included input files to get the html. Using `python build.py sphinx` would also do some post-processing and should give you the same as what is in the official docs except for the unavailable classes.

ghisvail:
> I could include just the generated documentation source files in the source tarball, but leave out the final .html files

I have added a commit which excludes the Sphinx build directory explicitly.

> you can add an extra step to run Sphinx on the included input files to get the html

This is what we do, hence me asking for the Sphinx *sources* to be included to the PyPI tarball.

RobinD42:
> hence me asking for the Sphinx *sources* to be included

Ah, okay. I didn't catch that before.  I'll make some tweaks to the build system later today or tomorrow.

RobinD42:
I added a few more commits and merged.

RobinD42:
See also #543

PR #454, opened 2017-08-07
wxget, wxdocs and wxdemo to scripts (#437)

### LINKED ISSUE #437: Suggest adding wxDocs and wxDemo to scripts
Currently for wxPython Phoenix the user has to locate the appropriate "Extras" package at https://sourceforge.net/projects/wxpython/files/wxPython4/extras/ download & unpack it.

This is a shame as the new documentation and demos package are invaluable _especially to new users_ 
 and is usually the first step in isolating potential bug reports, so making them hard to find is the last thing that is needed.

What I would like to suggest is adding to the Phoenix package two scripts to be automatically installed in the appropriate location:

 - wxDocs - when first run _possibly offer to_ download & unpack to a suitable location the docs package, subsequent runs _identified by the presence of the docs_ check if the docs version matches the current wx version, (if not offer to download), and open the docs at the root.  If the user declines the offline download direct them to the web documents.
 - wxDemo - If _the matching version_ demos package not available _offer to_ download and install to an appropriate location the demos package. Start the demos package.

_(As an aside while on the "Extras" page for wxPython 4 the user is prompted "Looking for the latest version? Download wxPython3.0-win64-**3.0.2.0**-py27.exe (16.8 MB)" or similar)_

GadgetSteve:
As discussed in #437 add new wxdemo & wxdoc launchers to scripts, these will ensure that the matching documents or demos for the loaded version of wx are available, downloading and installing them if required and then launch the  demo or the docs.

Both scripts are actually entry points for `wx/tools/wxget_docs_demo.py` which can be invoked directly with a parameter of either `docs` or `demo`.

To implement the fetching of the downloads under windows, which doesn't have wget, without introducing a dependency on `requests` it was necessary to implement a somewhat limited wxget which is usable as a library and as a command line script. This is implemented in `wx/tools/wxget.py`

Fixes #437


wx/tools/wxget_docs_demo.py
```diff
@@ -0,0 +1,175 @@
...
+
+import wx
+
+import wxget
+
+APP = None
+WXVERSION = wx.version().split(' ')[0]
+MAJOR = WXVERSION.split('.')[0]
+if MAJOR != '4':
+    raise ValueError("wx Versions other than 4 not supported!")
```
RobinD42: `wx.VERSION` is a tuple containing the version number components. Using it will save a couple lines here. Also, there is `wx.VERSION_STRING` for the string version of the full version number.

wx/tools/wxget_docs_demo.py
```diff
@@ -0,0 +1,175 @@
...
+
+def endsure_wx_app():
```
RobinD42: "endsure" --> "ensure" typo?

wx/tools/wxget_docs_demo.py
```diff
@@ -0,0 +1,175 @@
...
+def get_item(final, url, cache, name, ext):
+    """ Get the item """
+    print('Looking for', name, 'at', final)
+    fullpath = os.path.join(final, name)
+    if os.path.exists(fullpath):  # Already exists
+        return fullpath
+
+    cached = os.path.join(cache, name)
+    cached = os.path.extsep.join([cached, ext])
+    print('Looking for cashed', cached)
```
RobinD42: "cashed" --> "cached"

wx/tools/wxget_docs_demo.py
```diff
@@ -0,0 +1,175 @@
...
+    endsure_wx_app()
+    wx.MessageBox(err_text, caption='ERROR!',
+                  style=wx.OK|wx.CENTRE|wx.ICON_ERROR)
+
+def done(result=0):
+    """ Tidy up and exit."""
+    global APP
+    if APP:
+        print("Closing Launcher App!")  # Debug
+        APP.Destroy()
```
RobinD42: `wx.App.Destroy` can sometimes be unsafe to call, however since you are never calling `MainLoop` then it is sufficient to simply exit normally without cleaning up the app object. If you run into situations where that is not enough then using `wx.Exit()` to force it would be a better way to go.

wx/tools/wxget_docs_demo.py
```diff
@@ -0,0 +1,175 @@
...
+    pd = get_paths_dict()
+    location = get_item(pd['wxDocs'], pd['Docs_URL'], pd['Cache'],
+                        pd['Docs_Name'], pd['Ext'])
+    if location:
+        location = os.path.join(location, 'docs', 'html', 'index.html')
+        location_url = urlparse.urljoin('file:', pathname2url(location))
+        print("Show Docs at:", location)
+        webbrowser.open(location_url)
+    else:
+        result = 'Unable to find wx.Docs'
```
RobinD42: Replace "wx.Docs" with "wxPython Documentation" in the error message.

wx/tools/wxget_docs_demo.py
```diff
@@ -0,0 +1,175 @@
...
+    location = get_item(pd['wxDemo'], pd['Demo_URL'], pd['Cache'],
+                        pd['Demo_Name'], pd['Ext'])
+    if location:
+        cmds = [sys.executable, os.path.join(location, "demo", "demo.py")]
+        print("Launching", cmds[1])
+        pid = subprocess.Popen(cmds).pid
+        #subprocess.check_call(cmds) # Use instead for debug
+        print("Demo starting as PID %s - may take a few seconds!" % pid)
+    else:
+        result = 'Unable to find wx.Demo'
```
RobinD42: Replace "wx.Demo" with "wxPython Demo" in the error message

wx/tools/wxget.py
```diff
@@ -0,0 +1,168 @@
...
+    if not force:
+        dlg = wx.FileDialog(
+            None, message="Save As ...", defaultDir=dest_dir,
+            defaultFile=filename, style=wx.FD_SAVE | wx.FD_OVERWRITE_PROMPT
+        )
+        if dlg.ShowModal() == wx.ID_OK:
+            dest_dir, filename = os.path.split(dlg.GetPath())
+        else:
+            url = None
+        del dlg
```
RobinD42: `del dlg` is not enough, you need to actually call `dlg.Destroy()`. Or you can use a context manager, which will call `Destroy` for you on the way out, like this:

```python
    with wx.FileDialog(...) as dlg:
        if dlg.ShowModal() == ...
```

RobinD42:
Thanks!

In addition to the inline comments, please add a note about this to `CHANGES.txt`. 

It would probably also be a good idea to add something about these new scripts to `LONG_DESCRIPTION` in `setup.py`, so people who read the PyPI description will know about them. Similarly, adding something to `docs/sphinx/rest_substitutions/overviews/index.rst` will help people reading the docs online to fetch their own copy for local/off-line reading.

wx/tools/wxget_docs_demo.py
GadgetSteve: Done - much cleaner!

wx/tools/wxget_docs_demo.py
GadgetSteve: Done!

wx/tools/wxget_docs_demo.py
GadgetSteve: Changed to APP.Exit() - I have had a few cases where the main loop did not seem to be tidied up so thing that this will be safer.

wx/tools/wxget_docs_demo.py
GadgetSteve: Done

wx/tools/wxget_docs_demo.py
GadgetSteve: Done

wx/tools/wxget.py
GadgetSteve: Dpne

GadgetSteve:
Also added a yes/no prompt before going on-line for people who are on monitored or charged connections to have the option of saying NO!

wx/tools/wxget_docs_demo.py
GadgetSteve: Opps - wxExit not APP.Exit!

GadgetSteve:
Added these items to CHANGES.rst, (since wx/py/CHANGES.txt hasn't been updated since 0.9.7.9 (9/8/09)).

Added to the LONG_DESCRIPTION as suggested.

GadgetSteve:
Unable to get Sphinx to pick up these new modules correctly so simply added a note to `docs/sphinx/rest_substitutions/overviews/index.rst` - also added some handling of download errors, e.g. due to firewalls by giving the user the download URL and target location and suggesting using a browser.

PR #478, opened 2017-08-23
Change ToolBar.AddTool help parameters to be consistent

### LINKED ISSUE #475: wxToolBar.AddTool variants are differing for optional arguments (shortHelp vs. shortHelpString)

The method AddTool for a ToolBar currently expects different names for some optional arguments:

- shortHelp vs. shortHelpString
- longHelp vs. longHelpString

```python
# Put code sample here
toolbar.AddTool(wx.NewId(), "My Label", wx.Bitmap(32, 32), shortHelp="My short help")
toolbar.AddTool(wx.NewId(), "My Label", wx.Bitmap(32, 32), wx.Bitmap(32, 32), shortHelpString="My short help")
```

I would prefer using shortHelp and longHelp. I looked into the code and noticed these differences in wxWidgets too. Maybe it has to be modified there.

mesalu:
Change all *HelpString parameters in ToolBar.AddTool to be consistent internally and with wxPython Classic. 

Fixes #475

### DIFF
--- a/etg/toolbar.py
+++ b/etg/toolbar.py
@@ -91,7 +91,6 @@ def run():
             except ExtractorError:
                 pass
 
-
     c.find('OnLeftClick').ignore()
     c.find('OnMouseEnter').ignore()
     c.find('OnRightClick').ignore()
--- a/etg/toolbar.py
+++ b/etg/toolbar.py
@@ -100,14 +100,14 @@ def run():
 
     # Add some deprecated methods to aid with Classic compatibility.
     # TODO: Which others are commonly enough used that they should be here too?
-    c.addPyMethod('AddSimpleTool', '(self, toolId, bitmap, shortHelp="", longHelp="", isToggle=0)',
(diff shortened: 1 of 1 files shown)


mesalu:
Adding WIP label as change is ineffective in tests.

etg/toolbar.py
```diff
@@ -88,14 +100,14 @@ def _fixClientData(c):
 
     # Add some deprecated methods to aid with Classic compatibility.
     # TODO: Which others are commonly enough used that they should be here too?
-    c.addPyMethod('AddSimpleTool', '(self, toolId, bitmap, shortHelpString="", longHelpString="", isToggle=0)',
```
RobinD42: The `*SimpleTool` methods are only here for compatibility with Classic, which use the `*HelpString` version of the names (see https://github.com/wxWidgets/wxPython/blob/master/src/_toolbar.i#L217) so I think they should be left alone here.

mesalu:
Reverted the *SimpleTool changes.

Issue #463, opened 2017-08-09
Sorting issues in HyperTreeList 


ThomasP0815:
Operating system: Windows 7

wxPython version: 4.0.0a3

Stock or custom build: Stock

Python version: 3.6

Stock or custom build: Stock

Description of the problem:

I'm using a HyperTreeList in my project.
The function "SortChildren" is not working because an exception 
takes place in the class "CustomTreeCtrl".

Possible Patch - customtreectrl.py :

def SortChildren(self, item):
    ....

         -children = six.sort(children, self.OnCompareItems)
         -item._children = children
         +from functools import cmp_to_key
         +children.sort(key=cmp_to_key(self.OnCompareItems))

Thomas

PR #509, opened 2017-09-07
Fix for list.sort requiring keywords in python3.5

### LINKED ISSUE #508: Python3.5 list.sort requires keyword for arguments
**Operating system**: Windows (likely impacts all)
**wxPython version**: 4.0.0a3+
**Stock or custom build**: Stock
**Python version**: Python3.5+
**Stock or custom build**: Stock

Python3.5 changed list.sort (and sorted, but those are all correct) to use keyword only arguments.

bob-white:
Fixes #508

Python3.5 changed the parameters in list.sort to be keyword only arguments.
This fixes all the instances I could find.

### DIFF
--- a/wx/lib/agw/ribbon/bar.py
+++ b/wx/lib/agw/ribbon/bar.py
@@ -766,7 +766,7 @@ class RibbonBar(RibbonControl):
                         # Sneaky obj array trickery to not copy the tab descriptors
                         sorted_pages.append(info)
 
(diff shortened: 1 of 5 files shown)


RobinD42:
I think those are old "cmp" style functions and not the new "key" style functions. So you may need to use `functools.cmp_to_key` to get the correct sort orders like what was suggested in #463 and implemented in #507.  

https://docs.python.org/3/library/stdtypes.html?highlight=list%20sort#list.sort

Could you please test that and update the PR if needed?  Thanks!

PR #520, opened 2017-09-08
Add fedora-23 vagrant box

kollivier:
Right repo this time... :)

### DIFF
--- a/vagrant/fedora-23/Vagrantfile
+++ b/vagrant/fedora-23/Vagrantfile
@@ -1,0 +1,24 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure("2") do |config|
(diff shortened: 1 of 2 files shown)


swt2c:
Why Fedora 23 and not something still supported, like 26?

kollivier:
I have a project I'm working on where support for Fedora 23 is a hard requirement for now, and pip refused to install the Fedora 24 binary wheels when I tried, proceeding to install from source instead. So will Fedora 24 packages not work under 25 or 26? Adding support for new Vagrant boxes is pretty easy (kudos to Robin), but I'm not comfortable doing it for OS releases I won't be using or developing on.

swt2c:
Generally, no, packages for one Fedora release won't work on another because core library versions change.

PR #521, opened 2017-09-08
Sized controls null fix

kollivier:
This fix checks if the GetSizer() call returns None instead of a valid sizer, to avoid an error when using AUI.


RobinD42:
If you use separate branches for your PRs then this won't happen. 😃

kollivier:
Yup, I'm usually better than this, I swear! :P

PR #586, opened 2017-10-30
Allow listener functions to be type annotated

gopar:
If a listener would have type annotations, this would fail. 
It'll be nice if we could use type annons! 😄 

```python
from wx.lib.pubsub import pub

class Stuff:
    def __init__(self):
        pub.subscribe(self.test, 'hey')
        pub.sendMessage('hey', message='this is the message')

    def test(self, message: str) -> None:
        print(message)

Stuff()
```

### DIFF
--- a/wx/lib/pubsub/core/callables.py
+++ b/wx/lib/pubsub/core/callables.py
@@ -12,7 +12,7 @@ CallArgsInfo regarding its autoTopicArgName data member.
 
 """
 
-from inspect import getargspec, ismethod, isfunction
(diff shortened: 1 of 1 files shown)


RobinD42:
This should be submitted to the PyPubSub project as that is where our copy of the code comes from.  https://github.com/schollii/pypubsub 

That said, we are a few releases behind so unless Oliver objects I'll probably merge it here too.

RobinD42:
No longer relevant. See #792, #782

PR #603, opened 2017-11-08
10.13 OS X SDK fix

kollivier:
Make the min supported OS X version 10.6 to fix compilation issues with the OS X 10.13 SDK. The headers require weak-linking macro support that only exists on 10.6+.


RobinD42:
Sorry, I beat you to it:  https://github.com/wxWidgets/Phoenix/commit/09c066e695c2bddaf710ea11845c6e6637006ca3

😃

PR #665, opened 2018-01-05
Maintain reference to CallLater instances internally to prevent garbage collection

### LINKED ISSUE #457: wx.CallLater not always calling function
Operating system: windows 10 64 bit

wxPython version: 4.0.0.b1

Stock or custom build: stock

Python version: 3.6.2

Stock or custom build: stock

Description of the problem:
I've noticed on 2 of my applications that wx.CallLater will sometimes never call the function.  There's no error and its hard to reproduce reliably.  I've tried for quite a while to create some minimum code that reliably recreates the problem, but haven't had much luck so far.

mesalu:
Addresses #457, as of right now it is unclear if it resolves the initial issue.

Edit: Fixes #457


RobinD42:
> it is unclear if it resolves the initial issue.

I expect that it probably does.

PR #744, opened 2018-02-11
WIP: Switch Phoenix master to use wxWidgets master

RobinD42:
There is still more to be done, and some things still need fixed, but it seemed like a good time to make PR to record the current state of this work. 

Note that this will likely have build issues until my wxWidgets PR is merged, unless you also pull the changes in that PR into your local repo.  

https://github.com/wxWidgets/wxWidgets/pull/727


Metallicow:
@RobinD42 I never got an answer on the last pull request, but is the trailing whitespace for the rst substitutes necessary with the py files samples? Just noticed some sneaking back in... For example, to maintain a level of indentation...

RobinD42:
Not necessary, but it doesn't matter either way. That text originally is extracted from the wx interface xml files, has some simple substitutions and other manipulations performed on it, and then dumped out for the developer to deal with from there. Then they are merged in to the rest of the generated rst doc files, which are already crammed full of unnecessary whitespace. So in other words, IMO it's not worth your or my time to strip them out and create and merge a PR for it. A few tens or hundreds of extra bytes in files that are already tens of millions of bytes big is less than negligible.

Metallicow:
Thanks. I was just curious. Triming is usually just a last minute thing before doing a major RELz so there are not people complaining and making extra log noise when submitting bug fixes, shortly afterwards.

RobinD42:
Some Travis-CI builds timing out, but look like they would have been successful...

PR #764, opened 2018-02-22
Some new wxGtk widgets screenshots and a few wxMsw ones update

XPsoud:
I've just made some screenshots with wxGtk (Debian stretch with Mate)
I've also updated a few ones with wxMsw (Win.10)

I didn't test to rebuild the doc with these changes but :
- I've named the files regarding to [this page](https://docs.wxpython.org/DocstringsGuidelines.html)
- I've resized the images to correspond to the existing ones

Just let me know if I've made mistakes.

Regards
XAv'


mesalu:
I don't perceive there being issues if image size, and name, remain constant.
If you'd like to confirm your changes you can run: `build.py touch etg sphinx` to test your changes.
I recommend rebasing to wxPy-4.0.x first though since current master is... iffy. 
Another note: you'll need to install Graphviz (available via apt) and sphinx==1.5.1 to succesfully run the sphinx command in build.py


Just building the docs and inspecting a few of the html pages to ensure it doesn't cause weirdness therein should be sufficient.

Metallicow:
@XPsoud Running the png images thru FileOptimizer at max lossless, cuts down on load time with webpages. If you don't know how to do that, no biggie. I'm kinda surprised no one has ported it to a linux app yet...
Note to self: Check these when I do more screenshoting/image work.

mesalu:
Ran the generators locally and the output looked fine for the few pages I checked.

XPsoud:
@Metallicow As I've switched back to Windows to make the few Win10 screenshots, I passed the wxGtk related files to Riot, an Image Optimizer so their size has already been reduced. And the Win10 ones have been saved with Paint.net witch gives an already optimized result.

@mesalu I tried to build the docs ( under Windows using the make.bat script ) but I had an arror (Python can't find buildtools despite this package has been installed via pip)

RobinD42:
In this case buildtools is not a separate package but the folder in the source tree named `buildtools`. It has a `__init__.py` so it should be importable from scripts run in the parent folder unless you've got something funky going on with your `PYTHONPATH` or `sys.path`.

RobinD42:
> ( under Windows using the make.bat script )

I didn't notice this part before. YOu need to be in the root of the source tree and use the `build.py` script which has subcommands for building wxPython and the docs and other things. To build the docs you will also need to run the `etg` command to generate the ReST files for the wrapped classes. Some parts of the docs also need wxPython to be built but it may still be able to do the core without it. After `etg` you can run the `sphinx` subcommand to run sphinx.

mesalu:
To add on to @RobinD42  latest comment:
if your working directory is totally clean (e.g. fresh clone) you'll need to initialize the git submodule and run the`build.py dox` command.

That being said , I think the images are good. I'll go through my local doc build with a bit more attention to detail and report back later.

PR #782, opened 2018-03-12
Remove local pubsub 3.3.0 code and import 4.0.0 from external package

david-hughes:
Updates pubsub to latest version 4.0.0 by removing the local version 3.3.0 source code and importing version 4.0.0 from external stand-alone pubsub package (pip install Pypubsub).

This request arises out of discussion with pubsub author at https://github.com/schollii/pypubsub/issues/7 and the change permits wxPython code that uses pubsub to be frozen succssfully by PyInstaller. 

The inclusion of pubsub source code itself inside wx.lib is mainly historical and numerous other wx.lib items rely on external modules and packages being installed. It is being done this way in the interests of D.R.Y.


RobinD42:
Please add a ReST formatted docstring in `__init__.py` so we'll have something in the wxPython documentation about this. It doesn't have to be a lot, maybe just a few sentences about what pubsub is and why it's useful in a wx application. And also include a link to pubsub's docs, and give the pip command for installing it.

What do you think about adding it as a dependency in `setup.py`?

david-hughes:
I'm not familiar with ReST formatting but I can see how someone - Werner? - created the docstrings for wx.lib.pdfviewer and can mimic those. Is there a definitive guide to ReST formatting anywhere?

A dependency in setup.py like 'six'? Yes, why not. The extra disk space  - < 500k - is more than compensated for by the removal of version 3.3.0 in wx.lib

RobinD42:
There's some wxPython-specific info here, although it's mostly about documenting classes and methods: https://docs.wxpython.org/DocstringsGuidelines.html

There are some general intros and guides here: http://docutils.sourceforge.net/rst.html

mesalu:
Looks like you'll just need to add `'pypubsub'` to the `DEPENDENCIES' list on line 92 in /setup.py to accomplish the correct dependent behavior.
I'm definitely in favor of this if we're going to use it externally.

PR #844, opened 2018-05-05
Fix a bug in setting AuiDockingGuide size.

### LINKED ISSUE #727: agw.aui docking guide windows do not show correctly
I installed  4.0.0b2 gtk3 (phoenix) on ubuntu 16.04. When run the agw.aui example, the docking guide windows do not show correctly. 

It looks like when we set the size of the docking guide window (SetSize), the size will also include the size of the board and caption bar, which makes the 'client' area too small to show the docking guide correctly. 

gui

tianzhuqiao:
Set the client size to be the guide size. Otherwise, it is not shown correctly
on Linux.

Fixes #727

### DIFF
--- a/wx/lib/agw/aui/framemanager.py
+++ b/wx/lib/agw/aui/framemanager.py
@@ -2263,9 +2263,9 @@ class AuiSingleDockingGuide(AuiDockingGuide):
             else:
                 self.SetGuideShape()
 
-            self.SetSize(self.region.GetBox().GetSize())
+            self.SetClientSize(self.region.GetBox().GetSize())
         else:
-            self.SetSize((sizeX, sizeY))
+            self.SetClientSize((sizeX, sizeY))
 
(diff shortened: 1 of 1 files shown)


RobinD42:
Please retarget this to the `wxPy-4.0.x` branch and add an item to the `CHANGES.rst` mentioning the change and the issue number (#727). I'll cherry-pick it over to the master branch after it is merged.

RobinD42:
> Generally, what kind of change should be targeted to master, and what kind of change should be targeted to wxPy-4.0.x? 

Unfortunately there probably isn't a simple way to make that determination in most cases. Currently, if something is specifically about what's new in wxWidgets 3.1 then it should be targeted at master and everything else should go to 4.0.x. After 4.0.2 is released, then it will be  a little more fuzzy since there aren't definite plans yet for a 4.0.3 release and I may instead push forward to a 4.1.0 release (from the master branch.)

PR #849, opened 2018-05-11
Fixing listctrl (+ TextEditMixin) editor with height -1 in GTK

tfmoraes:
It's not possible to edit a listctrl cell value, because the editor has height -1. It's possible to see this error in the **ListCtrl_edit** from wxPython Demo. In the shell it shows this warning message:

`(demo.py:8618): Gtk-WARNING **: 15:34:15.183: Negative content height -1 (allocation 1, extents 1x1) while allocating gadget (node entry, owner GtkEntry)`

What I'm doing here to fix this error is calculating a text height (using self.editor.GetTextExtent('M')) and using this height values and editor height.

### DIFF
--- a/wx/lib/mixins/listctrl.py
+++ b/wx/lib/mixins/listctrl.py
@@ -612,7 +612,10 @@ class TextEditMixin:
         y0 = self.GetItemRect(row)[1]
 
         editor = self.editor
-        editor.SetSize(x0-scrolloffset,y0, x1,-1)
+        # It was using -1 as editor height, but it was not showing when in
+        # Linux. So it's calculating text height on the fly
+        height = self.editor.GetTextExtent('M')[1]
+        editor.SetSize(x0-scrolloffset,y0, x1, height)
 
         editor.SetValue(self.GetItem(row, col).GetText())
         editor.Show()


RobinD42:
I'm not sure that this is the right fix. The -1 was probably intended to indicate "don't change this value" or perhaps "use the default height", however since it is using the `(x,y,width,height)` version of `SetSize` then we may need to tell it how to interpret that -1 with the flags parameter. Please try it with `wx.SIZE_USE_EXISTING` or `wx.SIZE_AUTO` flag and see if one of those fixes the problem for you. 

The problem with using the text extent here is that it will only give you the height of the 'M' and won't include any internal margins or borders that the textctrl may be using.

If one of the flags above is not helpful then using `height = self.editor.GetBestSize().height` would probably be better than using the text extent.

tfmoraes:
Hi @RobinD42 ,

I've tested with all options you suggested me. I tested using ListCtrl_edit from demo.py. All of them makes GTK show this warning:
`
(demo.py:5530): Gtk-WARNING **: 12:28:30.138: Negative content height -5 (allocation 1, extents 3x3) while allocating gadget (node entry, owner GtkEntry)`

Using `height = self.editor.GetBestSize().height` the given height is 0 

Is there a way of get the internal borders and margins to add them to the GetTextExtent('M')?

Best regards.

RobinD42:
Fixed in a86b25b6b602ffa87760c36ef3a29d725bec44ad and 6453684bc3f27840650ea2ca2786a4a9828657f7

PR #852, opened 2018-05-12
Remove unnecessary notebook updating.

### LINKED ISSUE #686: AGW Aui Pane Header Drawing Bug
@infinity77 
While I have optimized and fixed bugs in my own xwx(Classic/Phoenix) custom copy of agw aui used for my SourceCoder, this is a bigger bug that I still haven't figured out how to fix yet and it is an obvious one that sticks out in the wxPy demo also. The best I can put it is it is a bug in the long looping process of drawing out agw aui interface.

**How to reproduce:**
Open the demo and start the agw 'AUI' demo. With the mouse grab one of the aui toolbars(I use a vertical one) by the gripper and start to move the toolbar up and down a bit in the available space without ripping it out of the user interface. You will notice the looping drawing steps of the aui auto notebook caption at the bottom(A third Auto-NB pane) flickering through all the pages captions. This seems to affect anything drawn as a pane header(icons, caption, min max close buttons, etc)

agwaui

### LINKED ISSUE #851: agw/AUI.py demo window flickering
On Linux (e.g., buntu 16.04) 
1. run the agw/aui.py demo,
2. turn on "Live Resize Update" option (in Options menu)
3. drag the gripper, the screen will flicker

tianzhuqiao:
Fixes #851, Fixes #686

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -93,6 +93,7 @@ Changes in this release include the following:
 
 * Fix a bug in setting AuiDockingGuide size. (#727)
 
+* Remove unnecessary notebook updating (wx.lib.agw.aui). (#851)
 
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -93,8 +93,9 @@ Changes in this release include the following:
 
 * Fix a bug in setting AuiDockingGuide size. (#727)
 
-* Remove unnecessary notebook updating (wx.lib.agw.aui). (#851)
+* Remove unnecessary notebook updating (wx.lib.agw.aui). (#851, #686)
 
+* Use BufferedDC in Repaint() to mitigate flicker (wx.lib.agw.aui). (#686)
 
 
 4.0.1 "Lemonade"
--- a/wx/lib/agw/aui/framemanager.py
+++ b/wx/lib/agw/aui/framemanager.py
@@ -6379,8 +6379,6 @@ class AuiManager(wx.EvtHandler):
         if not self.GetManagedWindow():
             return
 
-        if '__WXGTK__' in wx.PlatformInfo:
-            self.GetManagedWindow().Freeze()
         self._hover_button = None
         self._action_part = None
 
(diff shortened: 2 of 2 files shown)


Metallicow:
@RobinD42 @tianzhuqiao  This fixes #686 I have been trying to figure this out for ages. Thanks. Please test on linux flavors. Tested on Windows 7. Works like a charm.

Metallicow:
@tianzhuqiao  Please add #686 to the CHANGES.rst in another commit.

Metallicow:
After applying patch in SourceCoder xwx, the only thing I notice flickering that is left is the 'Custom Pane bitmap Buttons'. ... so there is another spot somewhere to address those to completely close #686. This occurs in the wxPy Demo also.

tianzhuqiao:
@Metallicow Thanks a lot for testing on Window 7. I tested it on Window 10, Mac OS 10.13, and Ubuntu 16.04 (and without the change, the demo on Ubuntu looks worst. Mac OS 10.13 looks good with/without the change.). 

For the 'Custom Pane bitmap Buttons' flickering, do you mean the 'max'/ 'min'/'close' button around the right corner of the pane? Have you tried replacing the following line (~line 8763 in framemanager.py, Repaint() function)?
    ```
    client_dc = wx.ClientDC(self._frame)
    ```
with
    ```
    if not self._frame.IsDoubleBuffered():
        client_dc = wx.BufferedDC(wx.ClientDC(self._frame), wx.Size(w, h))
    else:
        client_dc = wx.ClientDC(self._frame)
    ```

Metallicow:
@tianzhuqiao Yep, changing that line fixed the custom bitmap buttons(min/max/close) flicker. After looking at what you did, I just assumed it would be a loop issue, since the notebook tabs(order/drawing) when crazy. Could you add that as a commit also and that should successfully close both issues, unless there is any other visual abnormalities seen, testing on your end or another linux variants.

tianzhuqiao:
@Metallicow Thanks. Also tested on CentOS 1804; it looks good and similar to Ubuntu.

Metallicow:
Also add the other issue number to the opening post so it gets properly closed and referenced when this PullReqz gets merged. Other than that this looks good to me.

Metallicow:
@tianzhuqiao FYI as a general rule, anything in a previous version of Windows will work on latter versions, and since supposedly Win10 is "the last", XP would have been a better test. But we are in the middle of a new building and remodel or the refurb shop at the moment, so couldn't test. It may be awhile before I can test on XP again. Anywho.... Thank you very much for the work you performed; It is greatly appreciated. :)

PR #873, opened 2018-05-31
Fix a bug in group management on masked.numctrl

tizytissy:
Fixes #NNNN


RobinD42:
You should make this change in the same branch you used for your other PR, and then push the new commit. That will automatically add it to the original PR.

PR #828, opened 2018-04-27
Customtreectrl enhancement to have Window on CustomTreeItem on left of text

HelioGuilherme66:
Adds a the parameter `on_the_right`, defaulted True, to allow positioning the Window in a CustomTreeItem.
    
TODO: Make sure that Window is hidden when tree is collapsing. This is working at example of Proof of Concept at  `pip install https://github.com/HelioGuilherme66/RIDE/archive/gificons.zip` (file is src/robotide/ui/tree.py).

 ~~Sorry, I could not squash, avoid my master merges, and there are unwanted changes here.~~
@infinity77, @RobinD42 Please give me some guidance.
See this as a first proposal of an easy way to have animated GIF in CustomTreeItem. Thanks.

PR #886, opened 2018-06-17
Customtreectrl enhancement to have Window on CustomTreeItem on left of text

HelioGuilherme66:
Adds a the parameter `on_the_right`, defaulted True, to allow positioning the Window in a CustomTreeItem.

[x] - Updated Demo - To see these changes expand item **Hello World**
       - Added a Checkbox to set Window on the left (Unchecked by default)
       - To update the Window on the desired position (Play/Pause image) must enable TREE_ITEM_EXPANDING 
       - To hide these Windows must enable TREE_ITEM_COLLAPSING
       - To update the Window with the animated GIF (Recording image)  on the desired position must enable TREE_ITEM_CHECKED

This is the same as #828 but for wxPy-4.0.x


RobinD42:
Is this complete? The CustomTreeCtrl sample in the demo doesn't seem to be doing anything different when the option is changed. The Play/Pause images are always displayed on the right of the label. (At least on OSX.)

HelioGuilherme66:
Sorry @RobinD42, no it is not complete. I'll change the title with WIP prefix.

HelioGuilherme66:
@RobinD42 I consider this done now.

wx/lib/agw/customtreectrl.py
```diff
@@ -299,14 +299,14 @@ def __init__(self, parent):
 
 :class:`CustomTreeCtrl` is distributed under the wxPython license.
 
-Latest Revision: Andrea Gavana @ 09 Jan 2014, 23.00 GMT
+Latest Revision: Helio Guilherme @ 02 Jul 2018, 00.10 GMT
 
-Version 2.6
+Version 2.6.1
```
RobinD42: This is a significant new feature, so I would call it version 2.7.

PR #908, opened 2018-07-02
setup.py: list dependency on PIL used in a few wx.lib.agw modules

carandraug:
As discussed on IRC, here's a pull request that adds PIL as list of dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -90,6 +90,7 @@ Topic :: Software Development :: User Interfaces
 """
 
 DEPENDENCIES = [ 'six',
+                 'Pillow',
                  'PyPubSub']
 
 isWindows = sys.platform.startswith('win')


Metallicow:
@carandraug is this just for wx.lib.agw.shapedbutton? If so... I might look at the code and see if I can make a fix so it only imports if Pillow is installed.
My ShapedBitmapButton widget found here https://github.com/Metallicow/MCOW doesnt use PIL/Pillow.
For a moment I thought this was for thumbnailCtrl...

carandraug:
> is this just for wx.lib.agw.shapedbutton?

Yes. It is also an optional dependency of `wx/lib/agw/thumbnailctrl` and `wx/lib/agw/rulerctrl`

PR #960, opened 2018-08-13
Fix #957

### LINKED ISSUE #957: pycolourchooser palette click broken
**Operating system**: MacOS 10.13.6
**wxPython version**: 4.0.3 osx-cocoa (phoenix) wxWidgets 3.0.5
**Stock or custom build**: Stock, installed from conda
**Python version**: 2.7.15
**Stock or custom build**: Stock, miniconda distribution

I have an application that uses the PyColourChooser, including the palette. I've been working on updating it for wxpython4. I've found that when you click on the palette you get these error messages:

```
  File "/Users/jessehopkins/miniconda2/lib/python2.7/site-packages/wx/lib/colourchooser/pycolourchooser.py", line 464, in onPaletteDown
    self.doPaletteClick(event.GetX(), event.GetY())
  File "/Users/jessehopkins/miniconda2/lib/python2.7/site-packages/wx/lib/colourchooser/pycolourchooser.py", line 493, in doPaletteClick
    colour = self.getColourFromControls()
  File "/Users/jessehopkins/miniconda2/lib/python2.7/site-packages/wx/lib/colourchooser/pycolourchooser.py", line 384, in getColourFromControls
    return self.hsvToColour((h, s, v))
  File "/Users/jessehopkins/miniconda2/lib/python2.7/site-packages/wx/lib/colourchooser/pycolourchooser.py", line 370, in hsvToColour
    return wx.Colour(r, g, b)
TypeError: Colour(): arguments did not match any overloaded call:
  overload 1: too many arguments
  overload 2: argument 1 has unexpected type 'float'
  overload 3: argument 1 has unexpected type 'float'
  overload 4: argument 1 has unexpected type 'float'

  File "/Users/jessehopkins/miniconda2/lib/python2.7/site-packages/wx/lib/colourchooser/pycolourchooser.py", line 472, in onPaletteUp
    self.palette.ReleaseMouse() # Must call once for each CaputreMouse()
wxAssertionError: C++ assertion "!wxMouseCapture::stack.empty()" failed at /opt/concourse/worker/volumes/live/1bdcb0c4-b57f-443e-4aec-3572580f642b/volume/wxpython_1529592300425/work/ext/wxWidgets/src/common/wincmn.cpp(3319) in ReleaseMouse(): Releasing mouse capture but capture stack empty?
```

Upon investigation, it looks like in pycolourchooser.py, line 370 (in hsvToColour) is passing floats to wx.Colour instead of ints. This problem is easily duplicated just using the PyColourChooser in the wxpython demo.

I was able to fix it by simply changing line 370 to:

`
return wx.Colour(int(r), int(g), int(b))
`

This fixed both errors above. I'm hoping this can be easily incorporated into the next point release.

mesalu:
Fixes #957

### DIFF
--- a/wx/lib/colourchooser/pycolourchooser.py
+++ b/wx/lib/colourchooser/pycolourchooser.py
@@ -363,9 +363,11 @@ class PyColourChooser(wx.Panel):
         # Allow values to go full range from 0 to 255
         r, g, b = colorsys.hsv_to_rgb(hsv[0], hsv[1], hsv[2])
 
-        r *= 255.0
-        g *= 255.0
-        b *= 255.0
+        round_tenths = lambda x: int(x + 0.5)
+
+        r = round_tenths(r * 255.0)
+        g = round_tenths(g * 255.0)
+        b = round_tenths(b * 255.0)
 
         return wx.Colour(r, g, b)
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -62,7 +62,7 @@ Changes in this release include the following:
 * Switch wx.lib.plot to issue deprecation warnings with PlotPendingDeprecation
   so it doesn't have to enable all warnings to get them to be shown by default.
   (#902)
-  
+
 * Added a Python 3.7 builder on Fedora 28. (#925)
 
 * Fix the object ownership transfer for wx.Menu.Insert() (#931)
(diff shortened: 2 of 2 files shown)


wx/lib/colourchooser/pycolourchooser.py
```diff
@@ -363,9 +363,11 @@ def hsvToColour(self, hsv):
         # Allow values to go full range from 0 to 255
         r, g, b = colorsys.hsv_to_rgb(hsv[0], hsv[1], hsv[2])
 
-        r *= 255.0
-        g *= 255.0
-        b *= 255.0
+        round_tenths = lambda x: int(x + 0.5)
```
RobinD42: Is there a reason for not using the `round` builtin function?

mesalu: Dang, looks like I forgot to commit a comment explaining this...
Python2 will return a `float` from round even if the value is integral. (eg. `26.0` rather than `26`)

Since this change is supposed to ensure that `r`, `g`, and ,`b`are of type `int`, it seemed like a lambda expression to round tenths and ensure integral type was the better option.

PR #983, opened 2018-08-30
Double precision vector with scalar operand arithmetic

### LINKED ISSUE #971: Scalar multiplication and division for wx.Point and friends
**Operating system**: any
**wxPython version & source**: any
**Python version & source**: any

Multiplying a `wx.Point` by a double ends up truncating the floating point value before passing it to the C++ operator. This is likely just due to not including a wrapper for the `double` version of the operator.

Add the missing `double` operators for `wx.Point`, and others, and corresponding unit tests.  I think sip will be able to handle overloads between int/double, but that will need to be verified.

mesalu:
Fixes #971 

WIP, need to address ambiguous overload issues.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -99,6 +99,9 @@ Changes in this release include the following:
 
 * Fix type error that would occur using pycolourchooser. (#957)
 
+* Change scalar arithmetic on wx.Point, wx.RealPoint, and wx.Size
+  to truncate the result of the arithmetic, rather than the scalar
+  before arithmetic (#971)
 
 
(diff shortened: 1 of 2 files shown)


RobinD42:
It looks like there isn't an `operator/` with `double` arg for `wxPoint` and `wxSize`, so the compiler is trying to promote the point or size to an object that does have a `operator/` with `double` arg, (using implicit conversion ctors) but there is more than one of them to choose from. I don't know why there isn't a `double` version of `operator/`, but the fix for us would be to just leave the `int` version in the classes that don't have the double version of the operator.

PR #1035, opened 2018-10-12
include wxpy_api.h

### LINKED ISSUE #961: allow inclusion of wxpy_api.h by other projects
I'm currently working on getting Phoenix working with KiCad. To do this we currently need to include the file ```wxpy_api.h``` from the Phoenix source directory.

Would it be possible to add such headers into the Phoenix release packages?

RobinD42:
Fixes #961


Barracuda72:
It seems that although wxpy_api.h file is now included into the wheel, there is still a problem: wxpy_api.h uses SIP datatypes, but doesn't include sip.h header. I think this should be fixed.

PR #1038, opened 2018-10-15
fix optical grid appearance

jgehw:
- fixed optical appearance of row labels w.r.t. simple 3D effect
- fixed helpers for determining size (becomes important if you subclass a label renderer spanning multiple rows or columns)


jgehw:
checks failed with:
ImportError: /home/circleci/repo/wx/libwx_gtk3u_core-3.1.so.2: undefined symbol: SDL_OpenAudio

doesn't seem to be related with this pull request

PR #1079, opened 2018-11-13
Skip unimplemented tests instead of failing them

swt2c:


swt2c:
Changed to use xfail instead of skip so that these stand out as needing work still.

PR #1083, opened 2018-11-14
Fix pubsub provider test - restore expected whitespace

swt2c:


RobinD42:
FTR, I think the CI failure is due to branching from the 4.0.1 tag.

PR #1092, opened 2018-12-01
Clean up CircleCI dependencies

swt2c:
Since we're only building GTK+3, we don't need GTK+2 development packages.
Update to the most recent WebKit and SDL packages and add libmspack.


swt2c:
Pulled libmspack out - wxPython doesn't use it.

swt2c:
Updated the GL packages too.

PR #1103, opened 2018-12-08
Create wx.lib.asyncio to enable coroutine support

### LINKED ISSUE #1102: Integrate with the `asyncio` module for `async` and `await` coroutines support [WITH "PATCH"]
**Operating system**: Any
**wxPython version & source**: Latest PYPI
**Python version & source**: 3.5+

Coroutines are very helpful for GUI developing as this technique allows time-consuming tasks running along with the main event loop, while keep the program logic in a synchronous-like manner. `wx.Yield()` will no longer be necessary after the introduction of coroutines. A piece of sample code demonstrating the integration is attached in the end.

Because of the usefulness I think it will be great to add these helper classes into somewhere under the `wx.lib` package, while I have some questions. How should I submit the patch, is a pull request enough for a patch, and in which sub-package should these classes be located? Thank you very much!

```python3
#!/usr/bin/env python3
import asyncio
import concurrent.futures
import threading
import time
from asyncio.events import AbstractEventLoop
from asyncio.futures import Future
from typing import Optional, Callable, Any, Type

import wx


class WxTimerHandle(asyncio.TimerHandle):
    __slots__ = 'call_later',


class WxEventLoop(asyncio.AbstractEventLoop):
    def __init__(self, app: wx.AppConsole):
        self._closed = False
        self._app = app
        self._default_executor = None
        self._debug = False
        self._exception_handler = None
        self._task_factory = None

    def run_forever(self) -> None:
        self._app.MainLoop()

    def stop(self) -> None:
        self._app.ExitMainLoop()

    def is_running(self) -> bool:
        return self._app.GetMainLoop() is not None

    def close(self) -> None:
        executor = self._default_executor
        if executor is not None:
            self._default_executor = None
            executor.shutdown(wait=False)
        self._closed = True

    def _timer_handle_cancelled(self, handle: WxTimerHandle) -> None:
        handle.call_later.Stop()

    def call_soon(self, callback: Callable[..., Any], *args, context=None) -> None:
        self.call_soon_threadsafe(callback, *args)

    def call_at(self, when, callback: Callable[..., Any], *args, context=None) -> WxTimerHandle:
        return self.call_later(when - self.time(), callback, *args, context)

    def call_later(self, delay: float, callback: Callable[..., Any], *args: Any) -> WxTimerHandle:
        handle = WxTimerHandle(delay * 1000 + self.time(), callback, args, self)
        handle.call_later = wx.CallLater(int(delay * 1000), callback, *args)
        return handle

    def time(self) -> float:
        return time.monotonic()

    def create_future(self) -> asyncio.Future:
        return asyncio.Future(loop=self)

    def create_task(self, coro) -> asyncio.Task:
        if self._task_factory is None:
            return asyncio.Task(coro, loop=self)
        else:
            return self._task_factory(self, coro)

    def call_soon_threadsafe(self, callback: Callable[..., Any], *args, context=None) -> None:
        wx.CallAfter(callback, *args)

    def run_in_executor(self, executor: concurrent.futures.ThreadPoolExecutor, func: Callable[..., Any], *args) -> asyncio.Future:
        if executor is None:
            executor = self._default_executor
        if executor is None:
            executor = concurrent.futures.ThreadPoolExecutor()
            self._default_executor = executor
        return asyncio.wrap_future(executor.submit(func, *args), loop=self)

    def set_default_executor(self, executor: concurrent.futures.ThreadPoolExecutor) -> None:
        self._default_executor = executor

    def get_exception_handler(self):
        return self._exception_handler

    def set_exception_handler(self, handler):
        self._exception_handler = handler

    def default_exception_handler(self, context):
        print('Got exception: ' + repr(context))

    def call_exception_handler(self, context):
        if self._exception_handler is None:
            self.default_exception_handler(context)
        else:
            self._exception_handler(self, context)

    def get_debug(self) -> bool:
        return self._debug

    def set_debug(self, enabled: bool) -> None:
        self._debug = enabled

    def run_until_complete(self, future):
        raise NotImplementedError

    def is_closed(self) -> bool:
        return self._closed

    async def shutdown_asyncgens(self):
        raise NotImplementedError

    def set_task_factory(self, factory) -> None:
        self._task_factory = factory

    def get_task_factory(self):
        return self._task_factory


class WxEventLoopPolicy(asyncio.AbstractEventLoopPolicy):
    def __init__(self, app: Type[wx.AppConsole], delegate: asyncio.AbstractEventLoopPolicy = asyncio.get_event_loop_policy()):
        self._app = app
        self._loop = None
        self._delegate = delegate

    def get_event_loop(self) -> AbstractEventLoop:
        if threading.current_thread() is threading.main_thread():
            if self._loop is None:
                self._loop = WxEventLoop(self._app())
            return self._loop
        else:
            return self._delegate.get_event_loop()

    def set_event_loop(self, loop: AbstractEventLoop) -> None:
        self._delegate.set_event_loop(loop)

    def new_event_loop(self) -> AbstractEventLoop:
        return self._delegate.new_event_loop()

    def get_child_watcher(self) -> Any:
        return self._delegate.get_child_watcher()

    def set_child_watcher(self, watcher: Any) -> None:
        self._delegate.set_child_watcher(watcher)


def _bind_async(self, event, handler):
    def _handler(event):
        asyncio.ensure_future(handler(event))

    self.Bind(event, _handler)


wx.EvtHandler.BindAsync = _bind_async


def main():
    asyncio.set_event_loop_policy(WxEventLoopPolicy(app=wx.App))
    asyncio.get_event_loop().set_debug(True)

    def _another_loop_thread():
        nonlocal another_loop
        another_loop = asyncio.new_event_loop()
        asyncio.set_event_loop(another_loop)
        another_loop.run_forever()
        another_loop.close()

    another_loop = None
    another_loop_thread = threading.Thread(target=_another_loop_thread)
    another_loop_thread.start()

    async def on_close(event):
        def stop_and_close():
            another_loop.stop()

        another_loop.call_soon_threadsafe(stop_and_close)
        frame.Destroy()

    frame = wx.Frame(None, title='Coroutine Integration in wxPython', size=wx.Size(800, 600))
    frame.BindAsync(wx.EVT_CLOSE, on_close)
    frame.CreateStatusBar()
    frame.GetStatusBar().StatusText = 'Ready'
    counter = 1

    async def on_click(event):
        def log(message: str):
            frame.GetStatusBar().StatusText = message
            print(message)

        nonlocal counter
        count = ' [' + str(counter) + ']'
        counter += 1
        log('Starting the event handler' + count)
        await asyncio.sleep(1)  # Sleep in the current event loop
        log('Running in the thread pool' + count)
        # time.sleep is used to emulate synchronous time-consuming tasks
        await asyncio.get_event_loop().run_in_executor(None, time.sleep, 1)
        log('Running in another loop' + count)
        # Socket operations are theoretically unsupported in WxEventLoop
        # So a default event loop in a separate thread is sometime required
        # asyncio.sleep is used to emulate these asynchronous tasks
        await asyncio.wrap_future(asyncio.run_coroutine_threadsafe(asyncio.sleep(1), another_loop))
        log('Ready' + count)

    button = wx.Button(frame, label='\n'.join([
        'Click to start the asynchronous event handler',
        'The application will remain responsive while the handler is running',
        'Try click here multiple times to launch multiple coroutines',
        'These coroutines will not conflict each other as they are all in the same thread',
    ]))
    button.BindAsync(wx.EVT_BUTTON, on_click)

    frame.Show()

    asyncio.get_event_loop().run_forever()
    another_loop_thread.join()


if __name__ == '__main__':
    main()
```

gzxu:
(Actually this is a new feature) Fixes #1102

Adding `wx.BindAsync()` and other supporting classes to enable [coroutine](https://docs.python.org/3/library/asyncio-task.html#coroutine) support for wxPython.

Before the introduction of coroutines, developers for GUI applications have to ways to handle time-consuming tasks, periodically calling `wx.Yield()`, or making use of bare `wx.CallAfter()` with threads and callbacks. These methods are either inefficient or too complex, as described in the doc-string in this patch.

Coroutines are introduced to solve this problem. The program logic can be written in one or few coroutines, managing all the required states in the application, without worrying about blocking the main thread. Actually, coroutines are state-machines generated by the compiler (or exactly the Python interpreter). Additionally, as all coroutines (all coroutines launched in the main event loop, exactly) will be executed in the main thread, there will be no worry about any conflict caused by concurrency in the program logic.

### DIFF
--- a/wx/lib/asyncio.py
+++ b/wx/lib/asyncio.py
@@ -1,0 +1,336 @@
+#----------------------------------------------------------------------
+# Name:        wx.lib.asyncio
+# Purpose:     Coroutine support for event handlers
+#
+# Author:      XU Guang-zhao 徐广钊
+#
+# Created:     8-Dec-2018
+# Copyright:   (C) 2018 XU Guang-zhao 徐广钊
+# Licence:     wxWidgets license
+#
+# Tags:        documented
+#
+#----------------------------------------------------------------------
+
+"""
+Add ``wx.BindAsync()`` and other supporting classes to enable coroutine_
+support for wxPython.
+
+.. _coroutine: https://docs.python.org/3/library/asyncio-task.html#coroutine
+
+Description
+===========
+Before the introduction of coroutines, developers for GUI applications have to ways
+to handle time-consuming tasks. One of them is ``wx.Yield()`` which periodically
+yield the control to wxPython and let events to be processed. This is straitforward
+but performance issues are usually encountered, and users may frequently observe lags
+or busy cursors. Another one of them is to launch the task in another thread, however
+due to the complexity of GUI frameworks, UI elements can only be updated in the main
+thread, leading to the wide usage of ``wx.CallAfter()`` and callbacks. As the program
+logic increases, developers will eventually realize that they have already felt into
+the so-called "Callback Hell", and they will spend most of their time in taking care of
+the program states among these callbacks, because the program logic is usually in series
+but the program must be implemented in an asynchronous way. Considering such a situation:
+When a button is clicked a file will be downloaded, and then a CPU-intensive thread will
+be launched to process the data. With bare ``wx.CallAfter()`` there will be at least two
+callbacks, and what will happen if we need more, for example retrying the download when
+it fails?
+
+Coroutines are introduced to solve this problem. Time-consuming or IO-bound tasks can be
+wrapped into "awaitables" and ``await``-ed in coroutines; a coroutine will be
+automatically suspended when it is trapped in an "awaitable", and woke up when the task
+finishes. Additionally, as all coroutines (all coroutines launched in the main event
+loop, exactly) will be executed in the main thread, there will be no worry about any
+confilction caused by concurrency in the program logic.
+
+Usage
+=====
+Sample usage::
+    #!/usr/bin/env python3
+    import asyncio
+    import threading
+    import time
+
+    import wx
+    import wx.lib.asyncio
+
+    def main():
+        asyncio.get_event_loop().set_debug(True)
+
+        def _another_loop_thread():
+            nonlocal another_loop
+            another_loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(another_loop)
+            another_loop.run_forever()
+            another_loop.close()
+
+        another_loop = None
+        another_loop_thread = threading.Thread(target=_another_loop_thread)
+        another_loop_thread.start()
+
+        def on_close(event):
+            another_loop.call_soon_threadsafe(another_loop.stop)
+            frame.Destroy()
+
+        frame = wx.Frame(None, title='Coroutine Integration in wxPython', size=wx.Size(800, 600))
+        frame.Bind(wx.EVT_CLOSE, on_close)
+        frame.CreateStatusBar()
+        frame.GetStatusBar().StatusText = 'Ready'
+        counter = 1
+
+        async def on_click(event):
+            def log(message: str):
+                frame.GetStatusBar().StatusText = message
+                print(message)
+
+            nonlocal counter
+            count = ' [' + str(counter) + ']'
+            counter += 1
+            log('Starting the event handler' + count)
+            await asyncio.sleep(1)  # Sleep in the current event loop
+            log('Running in the thread pool' + count)
+            # time.sleep is used to emulate synchronous time-consuming tasks
+            await asyncio.get_event_loop().run_in_executor(None, time.sleep, 1)
+            log('Running in another loop' + count)
+            # Socket operations are theoretically unsupported in WxEventLoop
+            # So a default event loop in a separate thread is sometime required
+            # asyncio.sleep is used to emulate these asynchronous tasks
+            await asyncio.wrap_future(asyncio.run_coroutine_threadsafe(asyncio.sleep(1), another_loop))
+            log('Ready' + count)
+
+        button = wx.Button(frame, label='\n'.join([
+            'Click to start the asynchronous event handler',
+            'The application will remain responsive while the handler is running',
+            'Try click here multiple times to launch multiple coroutines',
+            'These coroutines will not conflict each other as they are all in the same thread',
+        ]))
+        button.BindAsync(wx.EVT_BUTTON, on_click)
+
+        frame.Show()
+
+        asyncio.get_event_loop().run_forever()
+        another_loop_thread.join()
+
+
+    if __name__ == '__main__':
+        asyncio.set_event_loop_policy(wx.lib.asyncio.WxEventLoopPolicy(app=wx.App))
+        main()
+
+"""
+
+#!/usr/bin/env python3
+import asyncio
+import concurrent.futures
+import threading
+import time
+from asyncio.events import AbstractEventLoop
+from asyncio.futures import Future
+from typing import Optional, Callable, Any, Type
+
+import wx
+
+
+class WxTimerHandle(asyncio.TimerHandle):
+    __slots__ = 'call_later',
+
+
+class WxEventLoop(asyncio.AbstractEventLoop):
+    def __init__(self, app: wx.AppConsole):
+        self._closed = False
+        self._app = app
+        self._default_executor = None
+        self._debug = False
+        self._exception_handler = None
+        self._task_factory = None
+
+    def run_forever(self) -> None:
+        self._app.MainLoop()
+
+    def stop(self) -> None:
+        self._app.ExitMainLoop()
+
+    def is_running(self) -> bool:
+        return self._app.GetMainLoop() is not None
+
+    def close(self) -> None:
+        executor = self._default_executor
+        if executor is not None:
+            self._default_executor = None
+            executor.shutdown(wait=False)
+        self._closed = True
+
+    def _timer_handle_cancelled(self, handle: WxTimerHandle) -> None:
+        handle.call_later.Stop()
+
+    def call_soon(self, callback: Callable[..., Any], *args, context=None) -> None:
+        self.call_soon_threadsafe(callback, *args)
+
+    def call_at(self, when, callback: Callable[..., Any], *args, context=None) -> WxTimerHandle:
+        return self.call_later(when - self.time(), callback, *args, context)
+
+    def call_later(self, delay: float, callback: Callable[..., Any], *args: Any) -> WxTimerHandle:
+        handle = WxTimerHandle(delay * 1000 + self.time(), callback, args, self)
+        handle.call_later = wx.CallLater(int(delay * 1000), callback, *args)
+        return handle
+
+    def time(self) -> float:
+        return time.monotonic()
+
+    def create_future(self) -> asyncio.Future:
+        return asyncio.Future(loop=self)
+
+    def create_task(self, coro) -> asyncio.Task:
+        if self._task_factory is None:
+            return asyncio.Task(coro, loop=self)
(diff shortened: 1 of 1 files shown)


wx/lib/asyncio.py
```diff
@@ -0,0 +1,336 @@
...
+
+"""
+Add ``wx.BindAsync()`` and other supporting classes to enable coroutine_
+support for wxPython.
+
+.. _coroutine: https://docs.python.org/3/library/asyncio-task.html#coroutine
+
+Description
+===========
+Before the introduction of coroutines, developers for GUI applications have to ways
```
RobinD42: to --> two

wx/lib/asyncio.py
```diff
@@ -0,0 +1,336 @@
...
+to handle time-consuming tasks. One of them is ``wx.Yield()`` which periodically
```
RobinD42: It's periodic only if you call wx.Yield periodically. You can just remove "periodically" here

wx/lib/asyncio.py
```diff
@@ -0,0 +1,336 @@
...
+yield the control to wxPython and let events to be processed. This is straitforward
```
RobinD42: yield --> yields

control to wxPython --> control to the event loop

wx/lib/asyncio.py
```diff
@@ -0,0 +1,336 @@
...
+        another_loop_thread.join()
+
+
+    if __name__ == '__main__':
+        asyncio.set_event_loop_policy(wx.lib.asyncio.WxEventLoopPolicy(app=wx.App))
+        main()
+
+"""
+
+#!/usr/bin/env python3
```
RobinD42: Remove this line. It only has effect as the first line in the file, and probably should not be in library modules anyway.

gzxu:
I am currently using it in my own proprietary projects, but I don't have time to write comments or documentation

sander76:
@gzxu @RobinD42  Want to add that I found wxasync .
Works like a charm !

leonardder:
I'd be also pretty interested in having this finished. The free open source screen reader NVDA relies on wx, and we won't be able to utilize asyncio effectively if there is no support for it within phoenix.

sander76:
> 
> 
> I'd be also pretty interested in having this finished. The free open source screen reader NVDA relies on wx, and we won't be able to utilize asyncio effectively if there is no support for it within phoenix.

I use wxasync now. Works perfectly. Would that be a solution for you ?

Metallicow:
If and when someone gets around to reviewing this...
In addition a demo should be made also and check for python versions, intentional oopsies, possible vulnerabilities(documentation could be one), etc...
Documenting this might be the hardest for some as it could be quite abstract, and while some folks are better coders, they may not be the best at documenting their code. So a few different use cases would be welcomed.

wx/lib/asyncio.py
Metallicow: Yes, the shebang should only reference itself on the firstline. Remove this otherwise you could have a possible security division/injection issue.
```if True: run()``` proves this as theta can be compiled into a "interpreter"

leonardder:
> 
> 
> > I'd be also pretty interested in having this finished. The free open source screen reader NVDA relies on wx, and we won't be able to utilize asyncio effectively if there is no support for it within phoenix.
> 
> I use wxasync now. Works perfectly. Would that be a solution for you ?

I'm not particularly happy with the gui message polling approach. It also introduces an extra dependency.

gzxu:
I'm not going to finish the remaining issues in a short time due to my lack of time. Would there be anyone continue on that? cc @RobinD42 @Metallicow

Metallicow:
@gzxu Most of our lives can be attributed to a short time. As far as mine or anyone elses, I cannot speak.
All I can say for now is that someone might pick it up in their spare time and finish it,
or someone will contact someone who thinks they know will finish it and maybe pay them to do so,
or someone real bright will show up to work drunk, steal the book, and the rest is history.
...and maybe send robin a tip or too...

Metallicow:
If 'you' was a human calculus experiment, maybe you would understand.

RobinD42:

https://discuss.wxpython.org/t/button-events-in-poll-mode/34217/9

leonardder:
I'm happy to try applying some of the requested changes. @gzxu What would be the best way to do so? I could of course file a pr against your fork, but given your lack of time, that might still be a problem for you?

RobinD42:
One way to do it is to start with @gzxu's  branch, add your changes, and then submit as a new PR with a note that links it to this PR.

leonardder:
> 
> 
> One way to do it is to start with @gzxu's branch, add your changes, and then submit as a new PR with a note that links it to this PR.

That makes sense. I forked phoenix and made a new branch. I've already applied the inline comments in https://github.com/leonardder/Phoenix/commit/be461ff771f76c51bc377ffe6ab3166b5bc68292 and will try to address the others later. First I need to dive into the world of asyncio again

leonardder:
I've applied most of the requests in [my asyncio branch](https://github.com/leonardder/Phoenix/tree/asyncio). Due to several reasons however, I'm reluctant to file this as a pr.

1. The process of creating a new event loop based on `asyncio.AbstractEventLoop` is poorly documented. I'm pretty sure that the current implementation of this branch only supports the basic subset of asyncio, no sockets, no pipes, etc. I've been looking at subclassing `asyncio.BaseEventLoop`, but at least the python 3.5 and 3.6 docs strongly discourage so as the API is unstable. However, `asyncio.BaseEventLoop` seems to implement some very fundamental stuff that's currently not in the WxEventLoop code. For example, `asyncio.events._set_running_loop` is called to set the currently running loop, but that seems to be py3.7 plus. Being compatible with multiple versions of Python as well as maintaining this code will therefore be pretty hard to do.
2. I'm a beginner in asynchronous programming with asyncio and really feel myself uncomfortable with taking responsibility for code that I'm unable to maintain or even unable to assess the full validity of.

My advise would be that this work either has to be continued bby a core dev, or not at all. It is just too complex for someone who isn't fully fused with WxPython development and asyncio.

I wonder whether @sirk390 would be interested in sharing is findings about his approach in wxasync. I assume he came up with his hybrid solution for a reason.

sirk390:
Yes, it's a complicated topic, and I also spend days navigating the source code to get my version to work.

The problem here is that this implementation doesn't support network, which is a considerable limitation (most programs that use asyncio require network). And I think it will be complicated to add and maintain. 

There would be a better approach, which is to patch the python asyncio event loop to detect GUI messages, and then wake up. I did this by patching "GetQueuedCompletionStatus" to call "MsgWaitForMultipleObjectsEx".
You can see my poc source code here: https://github.com/sirk390/wxasync/blob/master/test/poc_windows_patch_iocp.py (only works on windows)
However, after testing it, it was actually slower than my wxasync polling implementation. I then decided not to spend more time on it, but it could be interesting to know what goes wrong.

One small thing I didn't see in your implementation and that I like a lot, it to link the lifetime of a coroutine to a "wxWindow". That is, you cancel coroutines when the window is destroyed, and only allow BindAsync on wxWindow objects. This avoids a lot of manual work of canceling a Coroutine when you close the window, and avoids a lot of errors/warnings from asyncio or wxpython).
It also enforces better the principle of causality of Nathaniel Smith as explained here: https://vorpus.org/blog/some-thoughts-on-asynchronous-api-design-in-a-post-asyncawait-world/

Some other issues that are complicated (but most of the time not a big issue) are that the event loop can be frozen:
     * When you resize a window.
     * When you call some system modal dialogs. This could be solved by reimplementing these dialogs

PR #1128, opened 2019-01-07
Various bug fixes and feature additions to CustomTreeCtrl and HyperTreeList

cbeytas:
Mostly bug fixes and some feature additions.


RobinD42:
@infinity77, can you review the AGW parts of this PR? Thanks.

build.py
```diff
@@ -1181,7 +1181,12 @@ def injectClassInfo(className, srcTxt):
 def cmd_touch(options, args):
     cmdTimer = CommandTimer('touch')
     pwd = pushDir(phoenixDir())
-    runcmd('touch etg/*.py')
+    if isWindows:
+        # Windows touch
+        runcmd('cd etg && for %f in (*.py) do copy /B %f +,,')
+    else:
+        # Linux touch
+        runcmd('touch etg/*.py')
```
RobinD42: This could/should probably be done using pathlib2 on all platforms.

demo/Main.py
```diff
@@ -2697,6 +2697,7 @@ def main():
     except:
         pass
     app = MyApp(False)
+    locale = wx.Locale(wx.LANGUAGE_DEFAULT)
```
RobinD42: What problem did you have that necessitated this change?

cbeytas: I can't run demo.py in Idle under Windows/Python2.7 unless the locale is set.
Otherwise it throws the exception and the demo never runs. Full text:
```
wx._core.wxAssertionError: C++ assertion "strcmp(setlocale(LC_ALL, NULL), "C") == 0" failed at ..\..\src\common\intl.cpp(1579) in wxLocale::GetInfo(): You probably called setlocale() directly instead of using wxLocale and now there is a mismatch between C/C++ and Windows locale.
Things are going to break, please only change locale by creating wxLocale objects to avoid this!
```

infinity77:
> @infinity77, can you review the AGW parts of this PR? Thanks.

The changes are perfectly all right with me, I think @cbeytas has done a lot of good work on CustomTreeCtrl and HyperTreeList and I don’t see any blocker in merging this PR. I agree it’s enormous but it only impacts two AGW components.

PR #1205, opened 2019-04-15
Add use pip for osx and windows

ReblochonMasque:
use: ```pip install -U wxPython``` for osx and windows

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,6 +22,12 @@ platform specific code.
 .. contents:: **Contents**
 
 
+**NEW:** Installation with **pip** 
+----------------------------------
+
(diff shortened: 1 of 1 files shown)


ReblochonMasque:
I think the the project description on pypi should also show the `pip install` info. I could not find it to modify the file, but will do it if you indicate where it is located.

RobinD42:
The README is not really intended for people who will be using pip to do the install. I'll att a note prompting them to refer to the installation instructions on the website.

ReblochonMasque:
ok, thank you!

The readme, is in fact assuming that visitors already know what wxwidgets phoenix is all about, what state it is in, etc...

Now that it seems usable, and no longer "in transition" (?) maybe a clarification of the status would be a good thing?

PR #1215, opened 2019-04-28
Fixed scrollbar issue with ListCtrlAutoWidthMixin

RainingComputers:
Compensation for scrollbar width in ListCtrlAutoWidthMixin in GTK is not required. The compensation makes the widget smaller than it is supposed to be when it is made small enough for scrollbar to appear.


RobinD42:
Manually applied in 8b88b13c and 384cc827

PR #1227, opened 2019-05-18
Fix building with Python 3.8 on Linux

swt2c:
In Python 3.8, running "python3-config --libs" no longer returns the python
library itself.  This causes Waf's pyembed check to fail as it can't resolve
references to methods in the python library.  There is discussion[1] ongoing
about adding an "embed" mode to python3-config, but at the moment it is not
implemented.  We can work around this, however, in wxPython, by invoking Waf's
check_python_headers() with just the "pyext" feature as "pyembed" isn't needed
in wxPython (as far as I can tell).

[1] https://bugs.python.org/issue36721

### DIFF
--- a/wscript
+++ b/wscript
@@ -99,7 +99,7 @@ def configure(conf):
         # TODO: Check if it can/should be used on other platforms too.
         conf.my_check_python_headers()
     else:
-        conf.check_python_headers()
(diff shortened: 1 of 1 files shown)


RobinD42:
Thanks, I've been meaning to check if this is possible for some other reasons as well.

PR #1232, opened 2019-05-24
Remove pathlib2 as a build-time dependency for Python 3

cjmayo:
Built-in pathlib has provided the required touch() function since it was
introduced in Python 3.4:
https://docs.python.org/release/3.4.0/library/pathlib.html#pathlib.Path.touch

Fall back to pathlib2 for earlier versions of Python.

---

I've only tested this on Python 3.7.

### DIFF
--- a/build.py
+++ b/build.py
@@ -27,7 +27,10 @@ import tarfile
 import tempfile
 import datetime
 
-import pathlib2
+try:
+	import pathlib
+except ImportError:
+	import pathlib2 as pathlib
 
 from distutils.dep_util import newer, newer_group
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -27,7 +27,10 @@
 import tempfile
 import datetime
 
-import pathlib2
+try:
+	import pathlib
+except ImportError:
+	import pathlib2 as pathlib
```
RobinD42: Indent with 4 spaces.

RobinD42:
Superseded by #1241

PR #1246, opened 2019-05-30
Fixes #1235: CustomTreeCtrl edit label remains stuck forever

### LINKED ISSUE #1235: CustomTreeCtrl edit label remains stuck forever sometimes
Windows 10 (or Windows 7)
wxPython 4.0.6 (pip)
Python 3.7.3 [32bit]

From what I can tell, it seems that I'm able to somehow select the label from behind the TextCtrl and trigger a second edit label instance that can't be closed.
Or maybe KillFocus doesn't trigger correctly? I dunno.

Here's me reproducing it with the example from wxDemo:

Another example (from another GUI):

Nyral:
I added a condition to check that an edit control doesn't already exist.
For more info you can read [cbeytas's explanation](https://github.com/wxWidgets/Phoenix/issues/1235#issuecomment-496777121) about the issue.

### DIFF
--- a/wx/lib/agw/customtreectrl.py
+++ b/wx/lib/agw/customtreectrl.py
@@ -7733,8 +7733,9 @@ class CustomTreeCtrl(wx.ScrolledWindow):
         if self._editCtrl != None and item != self._editCtrl.item():
             self._editCtrl.StopEditing()
 
-        self._editCtrl = TreeTextCtrl(self, item=item)
-        self._editCtrl.SetFocus()
+        if self._editCtrl is None:
+            self._editCtrl = TreeTextCtrl(self, item=item)
+            self._editCtrl.SetFocus()
 
 
     def GetEditControl(self):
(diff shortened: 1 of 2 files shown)


wx/lib/agw/customtreectrl.py
```diff
@@ -7733,8 +7733,9 @@ def Edit(self, item):
         if self._editCtrl != None and item != self._editCtrl.item():
             self._editCtrl.StopEditing()
 
-        self._editCtrl = TreeTextCtrl(self, item=item)
-        self._editCtrl.SetFocus()
+        if self._editCtrl == None:
```
RobinD42: using `is` in place of `==` would probably be better

Nyral: That's how None comparisons are handled everywhere else in file.

Nyral: Done.

Nyral:
@RobinD42 Done.

Nyral:
Thanks!

PR #1284, opened 2019-06-29
The wx.WS_EX_VALIDATE_RECURSIVELY extended style flag is obsolete

### LINKED ISSUE #1278: sized_controls.SizedDialog failing with no WS_EX_VALIDATE_RECURSIVELY
**Operating system**: Windows 7  64-bit (also on MacOS Sierra)
**wxPython version & source**:   4.1.0a1.dev4130+927b6d94 from snapshot build
**Python version & source**:   Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)] on win32

`sized_controls.SizedDialog` seems to be broken in latest builds:

```python
#!/usr/bin/python

import wx
print wx.__version__

import wx.lib.sized_controls as sc

app = wx.App()
frame = wx.Frame(None, title='simple.py')
dlg = sc.SizedDialog(parent=frame, title="test dialog")
wx.StaticText(dlg.GetContentsPane(), label="a dialog")
frame.Show()
dlg.ShowModal()
app.MainLoop()
```
Returns this error:
```
  File "scdialog-broken-41.py", line 19, in <module>
    dlg = sc.SizedDialog(parent=frame, title="test dialog")
  File "u:/usr/extras/wx-4.1.0a1-win\wx\lib\sized_controls.py", line 600, in __init__
    self.SetExtraStyle(wx.WS_EX_VALIDATE_RECURSIVELY)
AttributeError: 'module' object has no attribute 'WS_EX_VALIDATE_RECURSIVELY'
```

RobinD42:
Fixes #1278

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -53,6 +53,11 @@ Other changes in this release:
   like wx.ListCtrl, wx.TreeCtrl and wx.dataview.DataViewCtrl. It has no effect 
   on the other platforms.
 
+* The wx.WS_EX_VALIDATE_RECURSIVELY extended style flag is obsolete, as it is
+  now the default (and only) behavior. The style flag has been added back into
+  wxPython for compatibility, but with a zero value. You can just stop using it
+  in your code with no change in behavior. (#1278)
+  
 
 
 
--- a/etg/defs.py
+++ b/etg/defs.py
@@ -108,6 +108,7 @@ def run():
     # TODO: these should be removed someday
     module.addPyCode("BG_STYLE_CUSTOM = BG_STYLE_PAINT")
     module.addPyCode("ADJUST_MINSIZE = 0")
+    module.addPyCode("WS_EX_VALIDATE_RECURSIVELY = 0")
 
 
     #-----------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


Metallicow:
@RobinD42 Is this planned to be removed at any point in the future?

Metallicow:
It should be a objective now to start documenting the closing/transition to python 3, and while some of these changes only seem to be in code or are subtle, there really needs to be an article or something documenting how we got here and possibly leave it in the public domain otherwise newcomers may be confused for a long time.

Two ways:
* Bite the bullet: Jump from py2 to py3+x
* Bit by Bit: Do it the way the bleeding edge developers do it. step by step

RobinD42:
> @RobinD42 Is this flag planned to be removed at any point in the future?
> ... and if so what timeframe.... wxPy 4.3, 4.5, etc.

Eventually, but given how things like this went in Classic that probably really means never. Since it's not a simple thing to be able to issue a DeprecationWarning when an integer variable is used it's hard to justify removing it entirely. It's been removed from the wxWidgets docs (which is what caused it to silently disappear from wxPython) so it should slowly fade away as new code is developed without it.

PR #1288, opened 2019-06-30
AddPrivateFont

Metallicow:
Demo example use of wx.Font.AddPrivateFont

Shows how to load a font.ttf at runtime that isn't installed on user's operating system.

Font Source:
https://github.com/adobe-fonts/source-code-pro


demo/data/SIL_OPEN_FONT_LICENSE.txt
```diff
@@ -0,0 +1,93 @@
+Copyright 2010, 2012 Adobe Systems Incorporated (http://www.adobe.com/), with Reserved Font Name 'Source'. All Rights Reserved. Source is a trademark of Adobe Systems Incorporated in the United States and/or other countries.
```
RobinD42: Let's add a line at the top of the file that states that this license applies to the SourceCodePro-Regular.ttf file.

demo/FontEnumerator.py
```diff
@@ -9,6 +17,11 @@ class TestPanel(wx.Panel):
     def __init__(self, parent, log):
         wx.Panel.__init__(self, parent, -1)
 
+        # Show how to add a private font to the application at runtime that
```
RobinD42: Instead of updating an existing sample please add a new one. This is a new feature so it should be highlighted with a new sample.

The new sample could be just a simple panel with a static text stating something like, "The font used in the text below was dynamically loaded from {filename}." followed by another static text using the font.

demo/data/SIL_OPEN_FONT_LICENSE.txt
Metallicow: Yea that might help a little more to resolve confusion as a oneline since SIL basically applies to the name of the font so there might not be a ton of mods to the thing in the future without stating that a certain char was changed, added etc, etc...

demo/FontEnumerator.py
Metallicow: Ok. will add another demo for this.

Metallicow:
@RobinD42 requested changes implemented. Does this look better?

demo/AddPrivateFont.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python
+
+import os
+import sys
+
+try:
+    gFileDir = os.path.dirname(os.path.abspath(__file__))
+except:
+    gFileDir = os.path.dirname(os.path.abspath(sys.argv[0]))
+gDataDir = gFileDir + os.sep + 'data'
```
RobinD42: please use `os.path.join` here

demo/AddPrivateFont.py
```diff
@@ -0,0 +1,115 @@
...
+
+#----------------------------------------------------------------------
+
+class TestPanel(wx.Panel):
+    def __init__(self, parent, log):
+        wx.Panel.__init__(self, parent, -1)
+
+        # Show how to add a private font to the application at runtime that
+        # doesn't have to be installed on the user's operating system.
+        wx.Font.AddPrivateFont(gDataDir + os.sep + 'SourceCodePro-Regular.ttf')
```
RobinD42: And here too.

Metallicow:
Ok. Guess this is ready now.

PR #1291, opened 2019-07-02
Update README.rst with linux requirements

tbenst:
libpng appears to be required? https://github.com/wxWidgets/Phoenix/blob/705aa63d75715f8abe484f4559a37cb6b09decb3/vagrant/centos-7/bootstrap.sh#L14

And `pyopengl` is optional? https://github.com/wxWidgets/Phoenix/blob/705aa63d75715f8abe484f4559a37cb6b09decb3/samples/glcanvas/GLCanvas.py#L160

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -355,6 +355,7 @@ or other unixes.
 * libnotify-dev
 * freeglut3
 * freeglut3-dev
+* libpng
 * libsm-dev
 * libgtk-3-dev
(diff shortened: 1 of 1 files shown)


RobinD42:
Updated the whole list in #1302 instead.

We don't mention other semi-dependent libraries in the README, so I didn't include that part.

PR #1301, opened 2019-07-09
Fix overlay crash

RobinD42:
Let wx.DCOverlay hold a reference to the dc to ensure it is not GC'ed first.


RobinD42:

https://discuss.wxpython.org/t/wxpython-4-1-0-released/34564/1

PR #1312, opened 2019-07-23
Fix typo in wx.StaticBoxSixer

driscollis:
This is just a fix to a typo in the docs that was mentioned in IRC


Metallicow:
Bravo. Brother. Bravo.

Metallicow:
@driscollis for six in one I do not fold, but for one in six i do not fold.

PR #1338, opened 2019-08-28
Feature: Colouring columns in hypertreelist

ThomasP0815:
Till now it is not possible to colour the columns in a hypertreelist 
(except the first).
I need this feature, because I want to highlight differencies between columns!

hypertreelist_coloured_columns

### DIFF
--- a/demo/agw/HyperTreeList.py
+++ b/demo/agw/HyperTreeList.py
@@ -1046,6 +1046,9 @@ class HyperTreeListDemo(wx.Frame):
         panel.SetSizer(sizer)
         sizer.Layout()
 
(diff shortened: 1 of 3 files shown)


cbeytas:
Can you also modify the demo so this feature can be tested?

ThomasP0815:
I modified the demo to support the feature mentioned above.

image

RobinD42:
I'm making a few additional changes, mostly to the sample, but also in the library module. Please double-check with your use-cases to make sure I didn't break anything.

Also, I noticed that if `TR_FULL_ROW_HIGHLIGHT` is turned on then the bg color from the first column is replicated across all columns even if they have their own bg colors set, which seems like a bug. I'll open a new issue for it. If you feel like making a PR for fixing that it would be appreciated.

PR #1359, opened 2019-09-14
Fix GetClippingRect to deal with the new GetClippingBox return value

RobinD42:
See https://discuss.wxpython.org/t/shifting-to-4-1-yes-really/32970/10


RobinD42:

https://discuss.wxpython.org/t/shifting-to-4-1-yes-really/32970/12

PR #1376, opened 2019-09-25
Fix numpy version to 1.16.1 for Python 2.7

### LINKED ISSUE #1374: Installation fails with Python 2.7 due to newest numpy
**Operating system**: All
**wxPython version & source**:          4.1.0a1
**Python version & source**:            Python 2.7.15

The latest numpy doesn't support Python 2.7, so pip fails installing the requirements for wxPython.
The version of numpy should be fixed for Python 2.7 to 1.16.1 since it is the latest version that supports Python 2.7. 
See: https://numpy.org/doc/1.16/release.html

As a workaround numpy 1.16.1 can be installed before wxPython manually.
```
pip install numpy==1.16.1
```

duohedron:
Fixes #1374

### DIFF
--- a/requirements/install.txt
+++ b/requirements/install.txt
@@ -1,4 +1,4 @@
 # Runtime dependencies needed when using wxPython Phoenix
-numpy
+numpy==1.16.1 ; python_version <= '2.7'
 pillow
 six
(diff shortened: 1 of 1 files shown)


RobinD42:
We should also add a line without the numpy version constraint for `python_version > 3`, correct?

duohedron:
> 
> 
> We should also add a line without the numpy version constraint for `python_version > 3`, correct?

Yes, you're right, I forgot it. I added it in the latest commit.

duohedron:
> 
> 
> May I ask why the dependency for numpy is fixed to 1.16.1?
> The latest numpy for Python 2.7 is 1.16.5

It was the latest version available by the time when I made the PR, and I only tested with that version. It can be changed to `numpy<1.17`

Issue #364, opened 2017-05-17
"Cannot find vcvarsall.bat" when installing via PIP on Windows


hyphz:
This error appears whenever I try to install via PIP on Windows, even when Visual Studio is installed and vcvarsall.bat is on the PATH.

wettenhj:
Are you trying to install a binary wheel?  Or are you trying to build from source?

If you are trying to build from source, you should specify the Visual Studio version you are using and the Python version you are using.  And the path to vcvarsall.bat might be relevant - from Command Prompt, you can run "where vcvarsall.bat".

RobinD42:
And also which version of Python you are using (and being used for running pip) and where that Python came from if it is not a stock build from python.org.

obdevel:
The problem manifests itself with more recent versions of Visual Studio (e.g. VS 2017), as the compiler version has been incremented to 10, whilst distutils seems to expect <= 9.

```C:\Users\xxx\wx\Phoenix>python setup.py build
running build
WARNING: Building this way assumes that all generated files have been
generated already.  If that is not the case then use build.py directly
to generate the source and perform the build stage.  You can use
--skip-build with the bdist_* or install commands to avoid this
message and the wxWidgets and Phoenix build steps in the future.

"C:\Python27\python.exe" -u build.py build
Will build using: "C:\Python27\python.exe"
2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)]
Python's architecture is 64bit
cfg.VERSION: 4.0.0a3

Running command: build
Running command: build_wx
Command '"C:\Python27\python.exe" -c "import distutils.msvc9compiler as msvc; mc = msvc.MSVCCompiler(); mc.initialize(); print(mc.cc)"' failed with exit code 1.
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Python27\lib\distutils\msvc9compiler.py", line 383, in initialize
    vc_env = query_vcvarsall(VERSION, plat_spec)
  File "C:\Python27\lib\distutils\msvc9compiler.py", line 271, in query_vcvarsall
    raise DistutilsPlatformError("Unable to find vcvarsall.bat")
distutils.errors.DistutilsPlatformError: Unable to find vcvarsall.bat
Finished command: build_wx (0.484s)
Finished command: build (0.484s)
Command '"C:\Python27\python.exe" -u build.py build' failed with exit code 1.```

[...]

PR #1424, opened 2019-10-31
Use setuptools' monkeypatching when getting supported compiler to build wx

leonardder:
Related to #364 

The current build system uses distutils to find an appropriate compiler. setuptools extends distutils to detect more compiler,s such as the VC2017 build tools. Setuptools also patches distutils in such a way that it gives much more detailed errors about what it misses, i.e. not just vcvarsall.bat couldn't be found, but Visual C++ 14.1 is missing, etc.

Filing this as a draft pr. I'm particularly interested in whether the tests will still pass for the several python versions supported.


leonardder:
Great, all tests pass, so I think this works as advertised.

PR #1468, opened 2019-12-22
wx.lib.gizmos.ledctrl text offset and text shift issue bugfix

kdschlosser:
This was caused by the use of GetClientSize being used when the control is constructed and then event.GetSize() when a size event took place. These 2 methods return different values and causes the text to shift. After this issue was solved another issue existed. there is a border around the control. this border is not directly specified to be drawn when the widget is rendered. the dc defaults to a white pen with a width of 1. a pen is not set before drawing the box. so the white pen there for gets used. This white border is not taken into consideration when the text is drawn and causes the text to be offset slightly. this corrects the offset issue. There is also no calculation in place to determine the actual width and actual height of each character and without having that it is more like a guess to figure out exactly where to draw the characters to have them be centered correctly inside of the box. I kept the same mechanism in place and made an adjustment to fix this issue i believe. I do think that the current way it is being done makes the code a tad more difficult to read and could be simplified. This is something I may take care of in a future PR.

There is no issue that has been made for this problem. 

Please do some tests to ensure that the repairs have been made correctly and are going to work properly.

### DIFF
--- a/wx/lib/gizmos/ledctrl.py
+++ b/wx/lib/gizmos/ledctrl.py
@@ -146,7 +146,7 @@ class LEDNumberCtrl(wx.Control):
         """
         if alignment != self.m_alignment:
             self.m_alignment = alignment
-            self._recalcInternals(self.GetClientSize())
+            self._recalcInternals()
 
             if redraw:
                 self.Refresh(False)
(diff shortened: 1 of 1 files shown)


kdschlosser:
TY Sir!

PR #1470, opened 2019-12-29
Fix wxget typo donwload and trim trailing space

Metallicow:
Fix wxget typo donwload and trim trailing space


RobinD42:
I'm applying part of this in #1474. The `_nanosvg.c` file is generated, so edits there will be overridden in the next build.

PR #1482, opened 2020-01-10
Cleanup per pep8 wxPy guidelines

Metallicow:
removes some unnecessary stuff, some evt to event readability changes, but mostly spaces around operator stuff and real minor stuff pycodestyle(formerly pep8.py) will complain about if run. Plus some of the files where real ugly about it since they hadn't been gone over in years and was completely inconsistant.


Metallicow:
Closing. Will try and work on a pycodestyle.py modified file to fit slight changes that would be affected otherwise send a newer PUL REQz based on the current master.

Metallicow:
small bits to consume is better than a bigger WOLLLLLLLLLLLMP all at once.

PR #1509, opened 2020-02-09
Fix ListCtrl Demo to work on wxPy4.0 and 4.1

Metallicow:
Don't break the demo on wxPy4.0 if hotswaping wxPy phoenix versions with wx.pth for example


swt2c:
No longer supporting wxPy 4.0, so closing this.

PR #1523, opened 2020-02-20
hotpatch time.clock -> time.perf_counter for py38 compatability

### LINKED ISSUE #1510: time.clock() removed in Python3.8 replace with time.perf_counter()
# time.clock() has been deprecated in py3.3 and removed in py3.8

https://docs.python.org/3/whatsnew/3.8.html
```
The function time.clock() has been removed, after having been deprecated since Python 3.3: 
use time.perf_counter() or time.process_time() instead, depending on your requirements, 
to have well-defined behavior. (Contributed by Matthias Bussonnier in bpo-36895.)
```

timeit test on py3

```python
def Test1():
    t = time.perf_counter()
def Test2():
    t = time.process_time()
```

```
Microsoft Windows 7 x64
Python 3.8.1 x64

------------------------------------------
Timeit Test - 1000000 times
Test1:[0.20428928500000154, 0.1874606649999997, 0.1864975280000003, 0.18765649500000237, 0.18984756899999766]
Test2:[0.45273342399999805, 0.45695258800000005, 0.45270879199999925, 0.4579686850000009, 0.45511786500000184]

Winner: Test1
Loser: Test2
```

And the alternative Py2 test just to see test results diff on 2/3 and x32/x64.

```python
def Test1():
    t = time.clock()
```

```
Microsoft Windows 7 x64
Python 2.7.17 x32

------------------------------------------
Timeit Test - 1000000 times
Test1:[0.1861580078093728, 0.16879239639197288, 0.1702609131900069]
```

It appears time.perf_counter() is the function we want to replace time.clock() with.

The purposed fix in framemanager.py:
Since python 2 is depricated now, 3 will become more popular, so hotpatch py2 since phoenix didn't run on less than py3.3 anyhow.

In framemanager.py imports section
```python
import time
import sys
if sys.version_info[0] == 2:  # hotpatch clock which is removed in py3.8
    time.perf_counter = time.clock
```

In framemanager.py AddPane1 method
```python
        # if the pane's name identifier is blank, create a random string
        if pinfo.name == "" or already_exists:
            pinfo.name = ("%s%08x%08x%08x") % (pinfo.window.GetName(), int(time.time()),
                                               int(time.perf_counter()), len(self._panes))
```

@RobinD42 I've tested this on 2and3 in my agw copy. Does this look good to you to submit a PullREQz?

Metallicow:
hotpatch time.clock -> time.perf_counter for py38 compatability

Fixes #1510

@RobinD42 Do I need to send a PullREQz to master also or will this update master also when merged?

### DIFF
--- a/wx/lib/agw/aui/framemanager.py
+++ b/wx/lib/agw/aui/framemanager.py
@@ -98,10 +98,10 @@ __date__ = "31 March 2009"
 
 import wx
 # just for isinstance
-try:
-    from time import time, process_time as clock
-except ImportError:
-    from time import time, clock
+import time
+import sys
+if sys.version_info[0] == 2:  # hotpatch clock which is removed in py3.8
+    time.perf_counter = time.clock
 import warnings
 
 import six
@@ -4754,8 +4754,8 @@ class AuiManager(wx.EvtHandler):
 
         # if the pane's name identifier is blank, create a random string
         if pinfo.name == "" or already_exists:
-            pinfo.name = ("%s%08x%08x%08x") % (pinfo.window.GetName(), int(time()),
-                                               int(clock()), len(self._panes))
+            pinfo.name = ("%s%08x%08x%08x") % (pinfo.window.GetName(), int(time.time()),
+                                               int(time.perf_counter()), len(self._panes))
 
         # set initial proportion (if not already set)
         if pinfo.dock_proportion == 0:


wx/lib/agw/aui/framemanager.py
```diff
@@ -98,10 +98,10 @@
...
 import wx
 # just for isinstance
-try:
-    from time import time, process_time as clock
-except ImportError:
-    from time import time, clock
+import time
+import sys
+if sys.version_info[0] == 2:  # hotpatch clock which is removed in py3.8
+    time.perf_counter = time.clock
```
RobinD42: Let's factor out the common bits and change the critical part of this to something like:

```
try:
    from time import perf_counter
except ImportError:
   from time import clock as perf_counter
```

I'd rather not modify a standard lib module if we don't need to.

RobinD42:
> Do I need to send a PullREQz to master also

Now that my focus is on 4.1.x releases my preference would be to have a PR for `master`. Then when I review it I'll add the `possible-backport` label if it's appropriate. That makes the the patch easy to find if there ever is a 4.0.8 release, and also for 3rd-party packagers that want to have newer fixes but not yet move on to 4.1.

RobinD42:
Closing in favor of #1524

PR #1534, opened 2020-02-27
Fixed a bug in the tabHeight calculation.

notgiven688:
flatnotebook:

The tabHeight gets calculated several times in flatnotebook.py. In the FNBRenderer class in "CalcTabHeight" the calculation is done differently.

### DIFF
--- a/wx/lib/agw/flatnotebook.py
+++ b/wx/lib/agw/flatnotebook.py
@@ -2197,15 +2197,11 @@ class FNBRenderer(object):
         dc = wx.MemoryDC()
         dc.SelectObject(wx.Bitmap(1,1))
 
-        # For GTK it seems that we must do this steps in order
-        # for the tabs will get the proper height on initialization
-        # on MSW, preforming these steps yields wierd results
         normalFont = wx.SystemSettings.GetFont(wx.SYS_DEFAULT_GUI_FONT)
         boldFont = normalFont
 
-        if "__WXGTK__" in wx.PlatformInfo:
-            boldFont.SetWeight(wx.FONTWEIGHT_BOLD)
-            dc.SetFont(boldFont)
+        boldFont.SetWeight(wx.FONTWEIGHT_BOLD)
+        dc.SetFont(boldFont)
 
         height = dc.GetCharHeight()
 


Metallicow:
That change looks a bit fishy... Do you have screenshots and any sort of specs for this change? What are you testing on with this? Linux/Windows? All it does is removes a GTK check and specific comment.

notgiven688:
I agree.

Below is a screenshot from Win10 (left: pull-request). Application is made highdpi aware with

```
import ctypes
try:
    ctypes.windll.shcore.SetProcessDpiAwareness(True)
except:
    pass
```

Windows Zoom is set to 150%:

compare

The extra GTK condition is absent in:
https://github.com/wxWidgets/Phoenix/blob/cbd3f712b057805c87efcffee74c752714ae836f/wx/lib/agw/flatnotebook.py#L1501

https://github.com/wxWidgets/Phoenix/blob/cbd3f712b057805c87efcffee74c752714ae836f/wx/lib/agw/flatnotebook.py#L2139

https://github.com/wxWidgets/Phoenix/blob/cbd3f712b057805c87efcffee74c752714ae836f/wx/lib/agw/flatnotebook.py#L3634

https://github.com/wxWidgets/Phoenix/blob/cbd3f712b057805c87efcffee74c752714ae836f/wx/lib/agw/flatnotebook.py#L5134

... it does exist in:

https://github.com/wxWidgets/Phoenix/blob/cbd3f712b057805c87efcffee74c752714ae836f/wx/lib/agw/flatnotebook.py#L2198

Metallicow:
Yea, that's what I figured you are using Win10. 
So basically that means the change you did isn't exactly right... That would break GTK.(If that bit of code is even still valid) I don't know who added that line for what specs(might have been ancient classic).
What needs done is to make some sort of highdpi check as a bool in the initialization, tho Im not sure if this would be an issue on mac or linux also.
@RobinD42  or @infinity77 might have a better answer as how to handle highdpi situations.
Some sort of GetIsHighDPI function but cross platform compatable...
This tho will need to be written better and tested on some other systems before applying.

notgiven688:
> So basically that means the change you did isn't exactly right... That would break GTK

Do not 100% agree, since the code path for GTK (in the pull request) is the same as before(?) I agree that there is something to fix and that someone with a deeper insight should check the code.

Thank you very much @Metallicow !

Metallicow:
Well stuff on retina displays seem to go weird sometimes and mac was the first ones out with the really high dpi stuff. Thats kinda what wreaks layout with images sometimes. Not sure exactly why. I normally test on normal settings as I don't always have access to the newest hardware. I on the other hand just got me a new raspberrypi4 other day need to setup but which kinda serves the opposite end of the spectrum so to speak Ill be testing lowDPI when I manage to get phoenix built on it. I'd guess it has to do with some sort of internal scaling factor tho or maybe a flag somewhere else that might need implemented. But no, I do not own a mac, so normally ask Robin or others to clarify if that stuff works usually.

Metallicow:
I cant seem to reproduce any weirdness on windows, and after looking at the wxPy2.8 stuff it appears only phoenix changes have occurred, so that leads me to believe after browsing though the source that comment might have related to possibly windows 2000 or something close...

```python
        normalFont = wx.SystemSettings.GetFont(wx.SYS_DEFAULT_GUI_FONT)
        boldFont = normalFont.MakeBold()
        dc.SetFont(boldFont)
``` 
This seems a bit more straight forward and could be a one-liner so, I'd go with this, if removing the commenting if no problems with mac.

... Also noticed that some of the check menuitems options in the demo file are reversed and need fixed up.

... Also I see this in the imports section of flatnotebook.py This is way old and classic is dead, so this should probably be removed also.
```python
# Check for the new method in 2.7 (not present in 2.6.3.3)
if wx.VERSION_STRING < "2.7":
    wx.Rect.Contains = lambda self, point: wx.Rect.Inside(self, point)
```

and I'd think that as far as the dpi, the hardcoded +=6 pixels should/might need to be scaled with whatever the percentage is however you might obtain that number...

... No comments by Andrea as to what OS's it was tested on in the docstring, but there is a obsolete/dead hyperlink there...
```python
# --------------------------------------------------------------------------- #
# FLATNOTEBOOK Widget wxPython IMPLEMENTATION
#
# Original C++ Code From Eran. You Can Find It At:
#
# http://wxforum.shadonet.com/viewtopic.php?t=5761&start=0
#
# License: wxWidgets license
```

notgiven688:
Great work @Metallicow 

To me it just looks like the Windows version does not get the correct font height since dc.SetFont is not called. The odd thing here is the strange comment which implies some deeper reasoning.

Issue #1555, opened 2020-03-20
Missing close for io.open, add close or use a with block in frame.py.


pbrod:
Replace code starting at:
https://github.com/wxWidgets/Phoenix/blob/ef1edacc205708cec7a0be4feb35cd524d8723bb/wx/py/frame.py#L926

with

```python
    with io.open(self.startupScript, 'r', newline=None, encoding='utf-8') as fid:
        text = fid.read()
```

pbrod:
Same issue occurs here as well:

https://github.com/wxWidgets/Phoenix/blob/ef1edacc205708cec7a0be4feb35cd524d8723bb/bin/mymd5.py#L9

Metallicow:
...as long as it is Py2/3 phoenix compatible, it looks good to me. @RobinD42 will be last person to verify/push anything thru.

PR #1561, opened 2020-03-20
Fixes issue #1555

pbrod:
Replace open-close statements by the use of "with"-blocks.

With the "With" statement, you get better syntax and exceptions handling. 

"The with statement simplifies exception handling by encapsulating common
preparation and cleanup tasks."

In addition, it will automatically close the file. The with statement provides
a way for ensuring that a clean-up is always used.


pbrod:
Closing due to errors

PR #1566, opened 2020-03-20
Use WX_CONFIG from the environment if set

### LINKED ISSUE #1488: [Request] don't hardcode wx-config
**Operating system**: Linux, FreeBSD
**wxPython version & source**:          
**Python version & source**:            

As a package builder we have to patch the setup to give the right variable to wx-config. Is it possible to allow the use of an environment variable to avoid this?

Ex: 
- Arch: https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/python-wxpython#n21
- Void: https://github.com/void-linux/void-packages/blob/master/srcpkgs/wxPython4/patches/0001-Switch-to-wx-config-gtk3.0-to-avoid-conflict-with-wx.patch#L29
And for FreeBSD (WIP), will be the same:
```
 	@${REINPLACE_CMD} -e "s|WX_CONFIG = 'wx-config'|WX_CONFIG = '${WX_CONFIG}'|" ${WRKSRC}/build.py
```
With wxPython3 we could add this variable: ```PYDISTUTILS_BUILDARGS=	WX_CONFIG="${WX_CONFIG}"```

Thank you

RobinD42:
if WX_CONFIG is set in the environment then use that instead of defaulting to "wx-config"	

Fixes #1488

### DIFF
--- a/build.py
+++ b/build.py
@@ -1573,13 +1573,16 @@ def cmd_build_py(options, args):
         os.environ['WXPYTHON_RELEASE'] = 'yes'
 
     if not isWindows:
-        WX_CONFIG = posixjoin(BUILD_DIR, 'wx-config')
-        if options.use_syswx:
-            wxcfg = posixjoin(options.prefix, 'bin', 'wx-config')
-            if options.prefix and os.path.exists(wxcfg):
-                WX_CONFIG = wxcfg
-            else:
-                WX_CONFIG = 'wx-config' # hope it is on the PATH
+        if 'WX_CONFIG' in os.environ:
+            WX_CONFIG = os.environ['WX_CONFIG']
+        else:
+            WX_CONFIG = posixjoin(BUILD_DIR, 'wx-config')
+            if options.use_syswx:
+                wxcfg = posixjoin(options.prefix, 'bin', 'wx-config')
+                if options.prefix and os.path.exists(wxcfg):
+                    WX_CONFIG = wxcfg
+                else:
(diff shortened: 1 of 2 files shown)


buildtools/config.py
```diff
@@ -400,7 +400,10 @@ def Verify_WX_CONFIG(self):
         """
         # if WX_CONFIG hasn't been set to an explicit value then construct one.
         if self.WX_CONFIG is None:
```
pbrod: I would propose to replace the if-else statement with the following:
self.WX_CONFIG = os.environ.get("WX_CONFIG",  "wx-config")

build.py
```diff
@@ -1573,13 +1573,16 @@ def cmd_build_py(options, args):
...
 
     if not isWindows:
-        WX_CONFIG = posixjoin(BUILD_DIR, 'wx-config')
-        if options.use_syswx:
-            wxcfg = posixjoin(options.prefix, 'bin', 'wx-config')
-            if options.prefix and os.path.exists(wxcfg):
-                WX_CONFIG = wxcfg
-            else:
-                WX_CONFIG = 'wx-config' # hope it is on the PATH
+        if 'WX_CONFIG' in os.environ:
```
pbrod: Replacing the if-else with the following will save one line:

```python
WX_CONFIG = os.environ.get("WX_CONFIG")
if WX_CONFIG is None:
     WX_CONFIG = posixjoin(BUILD_DIR, 'wx-config')
      .....
```

PR #1576, opened 2020-03-24
Fixes issue #1570:

### LINKED ISSUE #1570: Use of tempfile.mktemp is deprecated and poses a security risk.
The code starting at

https://github.com/wxWidgets/Phoenix/blob/8e2627e8e30d78a205080b8c2e337c748e9bb187/wx/tools/img2py.py#L153

could be replaced with  tempfile.NamedTemporaryFile(.....)

pbrod:
Replaced the unsafe "tfname = tempfile.mktemp()" call with
"with tempfile.NamedTemporaryFile(delete=False) as fid: tfname = fid.name"

* Fixes #1570

### DIFF
--- a/wx/tools/img2py.py
+++ b/wx/tools/img2py.py
@@ -34,7 +34,6 @@
 #
 
 
-
 """
 img2py.py  --  Convert an image to PNG format and embed it in a Python
                module with appropriate code so it can be loaded into
@@ -81,7 +80,6 @@ function. See its docstring for more info.
 
 import base64
 import getopt
-import glob
 import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


wx/tools/img2py.py
```diff
@@ -152,7 +148,9 @@ def img2py(image_file, python_file,
         app = wx.App(0)
 
     # convert the image file to a temporary file
-    tfname = tempfile.mktemp()
+    with tempfile.NamedTemporaryFile(delete=False) as fid:
+        tfname = fid.name
+
     try:
```
RobinD42: If you indent this block inside the `with` block, then you can also remove the deleting of the `tfname` file in the `finally` block and let `NamedTemporaryFile` take care of the cleanup.

pbrod: Well, it is not as simple as that.  NamedTemporaryFile() creates a file for writing and keeps it open until the "with block" ends. If I move the part inside the "try-finally block" inside the "with block" the fact that the file is still open will block the convert function to write to that file with the result that nothing will be converted.
For your idea to work the convert function must be refactored to also allow the fid to be passed in as an argument to convert function instead of the tfname string so that it can write to it directly (without opening it again since it is already open for writing).
However, the convert function is not so easy to refactor for the case when the image file is not the right type and you have to the wx.img2img conversion before saving.

So what do think?

RobinD42: Okay, it's alright the way it is then. I just thought it looked odd to still have to explicitly close the file and didn't think it through all the way.

I'll merge this in a little while.

PR #1579, opened 2020-03-25
Use generic animation classes on all platforms

### LINKED ISSUE #1567: wxAnimationDecoder is not available
**Operating system**: Windows 7 x64 SP1
**wxPython version & source**:    4.0.7.post1 msw (phoenix) wxWidgets 3.0.5 (pypi)
**Python version & source**: 3.7.5 Stackless 3.7 (tags/v3.7.5-slp:f7925f2a02, Oct 20 2019, 15:28:53) [MSC v.1916 64 bit (AMD64)]

I would like to be able to extend the animation portion of wxPython/wxWidgets to support APNG's I have found that wxAnimation has the method AddHandler but this method has not been added to wx.adv.Animation. In order for me to use that method I would need to subclass wxAnimationDecoder to pass to the method.

It appears to be fairly straight forward to be able to create a custom decoder and it would be nice if this was available to do instead of being limited to GIF and ANI. 

I am not sure if there might also be an issue because of wxAnimationType being used to identify the decoder being used.

If the ability to extend the animation portions of wx existed it would make it really easy to write GUI code that would handle security cameras and things of that nature. because the rendering of the images to the screen is handled in c code it would perform a lot better then wring a custom control to handle the animation in python. we just need a way to be able to add additional decoders.

```python
from wx.adv import AnimationDecoder
```

RobinD42:
Depends on: https://github.com/wxWidgets/wxWidgets/pull/1768
Fixes #1567 

The wxWidgets PR adds the ability to use the generic animation classes on all platforms, and also adds the generic-only methods and decoders to the interface headers.

This PR includes the changes needed to take advantage of all that.


RobinD42:
This issue has been mentioned on **Discuss wxPython**. There might be relevant details there:

https://discuss.wxpython.org/t/wxpython-4-1-0-released/34564/1

PR #1584, opened 2020-03-28
build.py: don't nuke PYTHONPATH

### LINKED ISSUE #1314: NixOS packaging misc.
Currently I'm working on packaging for NixOS again after someone else made some headway.
I've gotten tests to run, with some workarounds, as per https://github.com/NixOS/nixpkgs/pull/64047#issuecomment-515789547

Could you patch `build.py` to append to - instead of overwriting PYTHONPATH?
This is needed because `build.py` calls the test runner, which needs PYTHONPATH set correctly on NixOS, due to our use of nonstandard FHS.

If there's any other env vars treated similarly it might be good to fix them too.

deliciouslytyped:
Append phoenixDir to PYTHONPATH instead of overwriting it.

Partially Fixes #1314

### DIFF
--- a/build.py
+++ b/build.py
@@ -167,7 +167,7 @@ def main(args):
     setPythonVersion(args)
     setDevModeOptions(args)
 
-    os.environ['PYTHONPATH'] = phoenixDir()
+    os.environ['PYTHONPATH'] = os.environ['PYTHONPATH'] + ':' + phoenixDir()
     os.environ['PYTHONUNBUFFERED'] = 'yes'
     os.environ['WXWIN'] = wxDir()
 
@@ -2223,7 +2223,7 @@ def cmd_setpythonpath(options, args):
     assert os.getcwd() == phoenixDir()
 
     sys.path.insert(0, phoenixDir())
-    os.environ['PYTHONPATH'] = phoenixDir()
+    os.environ['PYTHONPATH'] = os.environ['PYTHONPATH'] + ':' + phoenixDir()
 
 
 


deliciouslytyped:
I think I also go it backwards because python seems to load from the first found result not the last one. I need to look into this more.

build.py
```diff
@@ -167,7 +167,7 @@ def main(args):
     setPythonVersion(args)
     setDevModeOptions(args)
 
-    os.environ['PYTHONPATH'] = phoenixDir()
+    os.environ['PYTHONPATH'] = os.environ['PYTHONPATH'] + ':' + phoenixDir()
```
RobinD42, line 170: `os.pathsep` should be used instead of ':'

build.py
```diff
@@ -2223,7 +2223,7 @@ def cmd_setpythonpath(options, args):
     assert os.getcwd() == phoenixDir()
 
     sys.path.insert(0, phoenixDir())
-    os.environ['PYTHONPATH'] = phoenixDir()
+    os.environ['PYTHONPATH'] = os.environ['PYTHONPATH'] + ':' + phoenixDir()
```
RobinD42, line 2226: Same here

RobinD42:
The `KeyError` in the CI builds are because `PYTHONPATH` is not already in  `os.environ`, so you need to either check if it's there already before trying to get it, or use something like `os.environ.get`

deliciouslytyped:
Thanks, will add the changes later!

evils:
@deliciouslytyped could i get you to take a look at this again?

deliciouslytyped:
"Its not that big of a deal" - but if you want it it'll probably be faster fr you to give it a shot because I haven't touched my wxPython stuff in a bit and I'm distracted by exams.

evils:
> I'm distracted by exams.

i'll submit a PR that closes this one if i get things building with nix, good luck

Issue #1465, opened 2019-12-20
Sequences/array with mixed types e.g. DataViewModel.GetValue


jensgoe:
**Operating system**: Win10
**wxPython version & source**:  4.1 (master), 4.0.7.post2
**Python version & source**: stock 3.7.3

**Description of the problem**:

After updating to wxPython 4.0.7.post2 i ran into TypeErrors "TypeError: Item at index 0 has type 'ModelItem' but a sequence of bytes or strings is expected"

I investigated the GetValue method of a DataViewModel implementation. It returned a tuple with an object and an integer. This seems to be not possible anymore because of #1355 
For me i worked around by returning a wrapper object that contains the tuple values. 

Because of the changed behavior i wanted to report this. The changed file arrays.sip looks like general array handling and the problem might occur in other places as well. Its up to you  am not sure if you want to treat it as a bug. Its up to you ;-). Maybe you have to add support for mixed type arrays?

Another solution i have in mind is to do an etg fix and redefine the return value of the DataViewModel.GetValue method to something else. 

```python
class MyModel(dv.PyDataViewModel):
    def GetValue(self, item, col):
        itemObj = self.ItemToObject(item)
        return itemObj, col # tuple with (object, int)

```

wossx:
Unfortunately, this is also a problem for me.
This is how I solve it temporarily:

```python
class CustomDataViewModel(dv.PyDataViewModel):

    def GetValue(self, item, col):
        ...
        self.temporaryValueTransfer = CustomClassWithProps()
        return self.temporaryValueTransfer           # or return ''


class CustomCellRenderer(dv.DataViewCustomRenderer):

    def SetValue(self, value):
        # print(value)          --> None

        value = self.GetView().GetModel().temporaryValueTransfer
        ...
        return True
```

Not nice, but until there is a solution at least it works.

jensgoe:
Behaviour has not changed since #1590. 
I noticed that the constructor argument varianttype of DataViewCustomRenderer defaults to "string" so returning a tuple in the model results in an type conflict. Passing "PyObject" as varianttype makes tuples work again. Maybe that has changed somewhere.

So the solution and probably safe way would be to set the right varianttype when using custom renderer:

```
class MyCustomRenderer(dv.DataViewCustomRenderer):
    def __init__(self):
        dv.DataViewCustomRenderer.__init__(self, "PyObject", mode=dv.DATAVIEW_CELL_EDITABLE)
```

PR #1590, opened 2020-04-07
Use different typecheck code for wxArrayInt and wxArrayString in variant conversions

### LINKED ISSUE #1581: SetValue in PGEditorDialogAdapter
Python 3.7.3
'4.0.7.post2 gtk3 (phoenix) wxWidgets 3.0.5'

Description: PGEditorDialogAdapter.SetValue() does not work for all objects.

Minimal example:

```python
import wx
import wx.propgrid as wxpg

class CustomClass():
    pass

class DA(wxpg.PGEditorDialogAdapter):

    def test(self):
        
        print("This works:")
        self.SetValue([1,2,3])
        print(self.GetValue())

        print("This does not:")
        self.SetValue(["1","2","3"])
        print(self.GetValue())

        print("-----------------")

        print("This works:")
        self.SetValue(CustomClass())
        print(self.GetValue())

        print("This does not:")
        self.SetValue([CustomClass(), CustomClass()])
        print(self.GetValue())

DA().test()
```

Produced output:

```
This works:
[1, 2, 3]
This does not:
None
-----------------
This works:
<__main__.CustomClass object at 0x7f09c53e9630>
This does not:
None
```

RobinD42:
Fixes #1581

### DIFF
--- a/src/arrays.sip
+++ b/src/arrays.sip
@@ -23,22 +23,8 @@
         // Code to test a PyObject for compatibility
         // Verify that the object is a sequence, but not bytes or unicode
         if (!sipIsErr) {
-            if (PySequence_Check(sipPy) &&
-                    !(PyBytes_Check(sipPy) || PyUnicode_Check(sipPy))) {
-
-                Py_ssize_t i, len = PySequence_Length(sipPy);
-                bool failed;
-                for (i=0; i<len; i++) {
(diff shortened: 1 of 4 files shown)


jensgoe:
Would this fix #1465 as well?

RobinD42:
Yes, I think so. Please test with the next dev snapshot if you are able and let me know.

PR #1620, opened 2020-05-01
ULC block item dragging

topic2k:
I wanted to block item dragging in UltimateListControl, but couldn't find a way. With this little addition i was able to get the desired result. If there is another way (any idea @infinity77 ?) , please tell me, otherwise i would love to see this PR merged.


RobinD42:

PR #1627, opened 2020-05-08
Fix issue #1364

### LINKED ISSUE #1364: aui toolbar not docking when it was just loaded as a floating pane. FIX included.
If a perspective is just loaded that contains an AuiToolbar in a floating pane, that toolbar cannot be docked back:

```
  File "/usr/local/lib/python3.7/dist-packages/wx/lib/agw/aui/framemanager.py", line 3217, in OnMoveEvent
    self.OnMoveStart(event)
  File "/usr/local/lib/python3.7/dist-packages/wx/lib/agw/aui/framemanager.py", line 3273, in OnMoveStart
    self._owner_mgr.OnMotion_DragToolbarPane(point)
  File "/usr/local/lib/python3.7/dist-packages/wx/lib/agw/aui/framemanager.py", line 9877, in OnMotion_DragToolbarPane
    ret, pane = self.DoDrop(self._docks, self._panes, pane, clientPt, self._action_offset)
AttributeError: 'AuiManager_DCP' object has no attribute '_action_offset'
```

This error is due to a bug in line `9877` of `wx/lib/agw/aui/framemanager.py` where the wrong variable `self._action_offset` is used. The correct variable to use in that line is `self.self._toolbar_action_offset`. In context, this is the correct call:
```
        if self.CanDockPanel(pane):
            # do the drop calculation
            ret, pane = self.DoDrop(self._docks, self._panes, pane, clientPt, self._toolbar_action_offset)
```

Once the correct variable is used, everything works as expected.

jmoraleda:
Fixes #1364

### DIFF
--- a/wx/lib/agw/aui/framemanager.py
+++ b/wx/lib/agw/aui/framemanager.py
@@ -9881,7 +9881,7 @@ class AuiManager(wx.EvtHandler):
         # is the pane dockable?
         if self.CanDockPanel(pane):
             # do the drop calculation
-            ret, pane = self.DoDrop(self._docks, self._panes, pane, clientPt, self._action_offset)
+            ret, pane = self.DoDrop(self._docks, self._panes, pane, clientPt, self._toolbar_action_offset)
 
         # update floating position
         if pane.IsFloating():
@@ -10751,5 +10751,3 @@ class AuiManager_DCP(AuiManager):
             # if we get here, there's no center pane, create our dummy
             if not self.hasDummyPane:
                 self._createDummyPane()
-
-


jmoraleda:
@RobinD42 What are your thoughts on this issue and my proposed fix?

Metallicow:
@jmoraleda Do you have a small running sample that can reproduce this on py2/3 and specs on the wxPython relz version? (no sample on the ticket page) I don't seem to have this issue with my program. In my code editor I can load from a settings.ini string or other methods. But I also at one point added oddball checks of sorts to make sure things was dockable on startup and not looking screwy. The agw aui sample doesn't save settings inbetween starts so it is hard to tell if this might be specific to something else. an ini sample should be fine to try and reproduce what you are getting i would think... not sure how you are trying to load the perspective...

PR #1628, opened 2020-05-10
new event wx.lib.agw.aui.EVT_AUI_PANE_CLOSED

topic2k:
I needed an event when a pane was closed, but the existing `wx.lib.agw.aui.EVT_AUI_PANE_CLOSE` is fired when a pane is about to be closed and that sometimes led to timing issues.
I added an extra event that is fired when the pane is really closed.
I was thinking about to rename EVT_AUI_PANE_CLOSE to EVT_AUI_PANE_CLOSING (which better fits), but that could break older code. So i just changed the docstring.

### DIFF
--- a/wx/lib/agw/aui/framemanager.py
+++ b/wx/lib/agw/aui/framemanager.py
@@ -132,6 +132,7 @@ if wx.Platform == "__WXMSW__":
 # AUI Events
 wxEVT_AUI_PANE_BUTTON = wx.NewEventType()
 wxEVT_AUI_PANE_CLOSE = wx.NewEventType()
+wxEVT_AUI_PANE_CLOSED = wx.NewEventType()
 wxEVT_AUI_PANE_MAXIMIZE = wx.NewEventType()
 wxEVT_AUI_PANE_RESTORE = wx.NewEventType()
 wxEVT_AUI_RENDER = wx.NewEventType()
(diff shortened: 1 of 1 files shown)


wx/lib/agw/aui/framemanager.py
```diff
@@ -5080,6 +5083,13 @@ def ClosePane(self, pane_info):
         if to_destroy:
             to_destroy.Destroy()
 
+        # Now inform the app that we closed a pane.
+        e = AuiManagerEvent(wxEVT_AUI_PANE_CLOSED)
+        e.SetPane(pane_info)
+        e.SetCanVeto(False)
+        e.SetId(wxEVT_AUI_PANE_CLOSED)
```
RobinD42, line 5090: An Event's id attribute  is normally used for the id of the widget sending the event, not the event type. For example, for a `EVT_BUTTON` event the id is the value of the button's `GetId()`, not `wxEVT_BUTTON`. Also, it looks like the other AUI events are not setting the id (not great, but okay) so it's probably better to follow the existing pattern here.

wx/lib/agw/aui/framemanager.py
```diff
@@ -5080,6 +5083,13 @@ def ClosePane(self, pane_info):
...
+        wx.CallAfter(self.ProcessMgrEvent, e)
```
RobinD42, line 5091: Is wx.CallAfter necessary here?

RobinD42: Also, would it make sense to just use the `FireEvent` method to create and send the event instead of doing it all here?

wx/lib/agw/aui/framemanager.py
topic2k: I see why i set the id: to distingish from other events. But now i know i can use the event type instead.

wx/lib/agw/aui/framemanager.py
topic2k: Not 100% sure, but i think i thought to have the closing event be finished before fireing the closed event. Anyway, i switch to `FireEvent` which in also uses `wx.CallLater`.

topic2k:
@RobinD42 what is preferred: just adding a commit with changes or squashing the commits and force-push?

RobinD42:
Just adding the new commits is fine. It makes it easy to see what's changed. If a PR gets messy enough that a squash would be better, then I can do that when merging.

RobinD42:

PR #1630, opened 2020-05-10
[ULC] add parameter column index to SendNotify

topic2k:
Added a column parameter to SendNotify and set it to the column index in an UltimateListEvent where applicable.


RobinD42:

PR #1650, opened 2020-05-27
Set transfer flag for subMenu in MenuItem constructor

### LINKED ISSUE #1648: Crash when creating submenus
**Operating system**: Windows 10 or Mac OSX
**wxPython version & source**:          4.1.0 binary from PyPI
**Python version & source**:             2.7.12

If one tries to create a submenu with a wx.MenuItem, it crashes the application with a segmentation fault.  Happens if you just right click to pop the menu, and then click off to dismiss the menu.  To demonstrate, change the following line in the PopupMenu.py wxPython demo.  This is right around line 96.
[PopupMenu_bug.zip](https://github.com/wxWidgets/Phoenix/files/4672614/PopupMenu_bug.zip)

```python
        #menu.Append(self.popupID7, "Test Submenu", sm) # original line of code
        sm = wx.MenuItem(menu, self.popupID7, "Test Submenu", subMenu=sm) 
        menu.Append(sm)        
```

Doing it that way had always worked in wxPython 3.  Attached is the full demo code that crashes.

swt2c:
This fixes a crash when wx.MenuItem with a submenu.

Fixes #1648.

### DIFF
--- a/etg/menuitem.py
+++ b/etg/menuitem.py
@@ -36,6 +36,7 @@ def run():
     c.addPrivateCopyCtor()
     tools.removeVirtuals(c)
 
+    c.find('wxMenuItem.subMenu').transfer = True
     c.find('SetSubMenu.menu').transfer = True
 
     # deprecated and removed


swt2c:
I'd also like to request this be backported to the 4.0 branch in case there's ever another release there.

Issue #1636, opened 2020-05-13
PropertyGrid demo (custom PGEditors) crashing wxPython 4.1.0


molokov:
**Operating system**: Windows 10 64-bit, CentOS7 x64
**wxPython version & source**:   4.1.0, built from source
**Python version & source**: Python 2.7.12, self built 

**Description of the problem**:
Several of the "Custom Properties and Editors" of the PropertyGrid demo (demo/PropertyGrid.py) are not working with wxPython 4.1.0, causing a crash of the application. 

Many of these examples are also not up to date with the latest wxWidgets used by this release.
* wxPG_SUBID1 and wxPG_SUBID2 are no longer used by wxPGEditor, the controls just use wxID_ANY in propgrid/editors.cpp
* wxPGWindowList.m_primary is no longer available, but there is a GetPrimary() method instead.

In the PropertyGrid.py demo, I have made substitutions to get past these wxWidgets changes (wxpg.PG_SUBID1 / PG_SUBID2 -> wx.ID_ANY; m_primary -> GetPrimary()), but then the examples will cause the application to crash.

Using a debug build of wxPython, I see the following assertion:

    python: ../../../../sip/siplib/siplib.c:3444: parseResult: Assertion `assign_helper != ((void *)0)' failed.
Aborted (core dumped)

This assert fails when trying to edit the following rows of the demo:
* ImageFileWithLargeEditor
* IntProperty2 
* MultipleButtons

(There is also an unrelated exception when the 'SetPropertyValues' or 'GetPropertyValues' or 'GetPropertyValues (as_strings=True)' or 'AutoFill' buttons are pressed)

--
My application (working with wxPython 4.0.7post2) makes use of custom editors, based on the SampleMultiButtonEditor from this demo - but given that this is no longer working with the 4.1.0 release, it would be nice to know if the demo just needs to be updated, or if the changes in wxWidgets require a fix to wxPython's SIP code so that we can still have custom PGEditor classes in Python.

swt2c:
I am looking into this.  I see the problem - it's that sip is trying to copy an instance of wxPGWindowList and it can't do that.  It seems to be related to the removal of the default constructor that was in wxWidgets 3.0, as this worked there.

RobinD42:
Closed by #1687

PR #1658, opened 2020-06-04
Fix some issues with PropertyGrid demo

swt2c:
This PR fixes some issues with the PropertyGrid demo as reported in #1636.  Additionally, to fix the assertion problem reported there, a new version of SIP is required.  For that, some additional private assignment operators are required to provide correct hints to SIP.


swt2c:
BTW, you don't need to wait until you have the new SIP in place to merge this.  The private assignment operators won't hurt anything with the existing SIP.

RobinD42:
Ok, thanks for the clarification.

PR #1672, opened 2020-06-08
Optimize auitoolbar Realize method loop

Metallicow:
Optimize auitoolbar Realize method loop. take some of the lookups out of the loop.

### DIFF
--- a/wx/lib/agw/aui/auibar.py
+++ b/wx/lib/agw/aui/auibar.py
@@ -2993,6 +2993,8 @@ class AuiToolBar(wx.Control):
 
         # create the new sizer to add toolbar elements to
         sizer = wx.BoxSizer((horizontal and [wx.HORIZONTAL] or [wx.VERTICAL])[0])
+        # local opts
+        sizer_Add, sizer_AddSpacer, sizer_AddStretchSpacer, sizer_SetItemMinSize = sizer.Add, sizer.AddSpacer, sizer.AddStretchSpacer, sizer.SetItemMinSize
 
         # add gripper area
         separator_size = self._art.GetElementSize(AUI_TBART_SEPARATOR_SIZE)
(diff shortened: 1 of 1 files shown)


RobinD42:
Did you profile this before and after your changes? Just curious how much of a change it would make on modern hardware. My gut feel is that it would take very many iterations (thousands) before changing attribute access to local variables would make a noticable difference to a human, but I suppose I could be wrong.

Metallicow:
Most of the individual method optimizations speed it up just a little bit, but the python aui sure does quite a bit of redrawing when certain events always happen. Overall it it small per function call but when the methods get called multiple times on startup and or when in runtime even .01 to .1 does add up to about a second for me anyway. I got around 10-20 toolbars that change at times. sometimes I just need to swap a icon for a disabled version and update it etc. ex: searchctrl, cut copy paste tools etc.
I usually just use timeit on the individual copied methods to do the benchmark live at runtime and see if there is even a minimal difference with the changes, then measure it with multiple widgets.

Metallicow:
basically with all the local opts and loops being optimized in my app on wxPy4.0 it drops startup time from like 5-6 seconds down to 2, which is nice. Usually perspectives and work space changes are a big part of the startup that calls the drawing stuff. Plus I tend to restart my program alot to test changes inbetween newer wxPy versions so I got xwx compatible code.

Metallicow:
@RobinD42 Just since you was curious, most of the on the fly testing is done on a laptop.
Of course I got access to various eras of hardware, but this I believe is acceptable specs for nowadays.
Not the greatest and not crap either.
Intel Core i7 4710HQ @ 2.50GHz Haswell 22nm Technology
16.0GB Dual-Channel DDR3 (11-11-11-28)
NVIDIA GeForce GTX 970M

RobinD42:
Ok, thanks for the details.

Metallicow:
I'm not saying that all optimizations will run faster on all systems, but if you do things multiple times(many times/recursive) you will notice a difference. For some folk, it is a blink of an eye, for others they can notice it. In C++ the same people would blink and eye and say 'Huh. didn't see a thing.'
Python optimizations are not the same as doing it in C.

PR #1674, opened 2020-06-08
Optimize auitoolbar OnPaint method loop

Metallicow:
Optimize auitoolbar OnPaint method loop. reduce the amount of lookups in the loop.

### DIFF
--- a/wx/lib/agw/aui/auibar.py
+++ b/wx/lib/agw/aui/auibar.py
@@ -3482,6 +3482,9 @@ class AuiToolBar(wx.Control):
             last_extent -= dropdown_size
 
         # paint each individual tool
+        # local opts
+        _art = self._art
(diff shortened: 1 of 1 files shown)


wx/lib/agw/aui/auibar.py
```diff
@@ -3482,6 +3482,9 @@ def OnPaint(self, event):
             last_extent -= dropdown_size
 
         # paint each individual tool
+        # local opts
+        _art = self._art
+        DrawSeparator, DrawLabel, DrawButton, DrawDropDownButton, DrawControlLabel = _art.DrawSeparator, _art.DrawLabel, _art.DrawButton, _art.DrawDropDownButton, _art.DrawControlLabel
```
RobinD42, line 3487: Let's avoid extra long lines like this.

Metallicow: Ok. I wrapped the line.

PR #1679, opened 2020-06-10
pywxrc: fix embedded binary resources for python3

jeremyd2019:
Previously, I was getting errors 'No handler found for image type' and 'Unknown image data format' after updating from wxPython 'classic'.

### DIFF
--- a/wx/tools/pywxrc.py
+++ b/wx/tools/pywxrc.py
@@ -216,7 +216,7 @@ def __init_resources():
     __res.Load('%(resourceFilename)s')"""
 
     FILE_AS_STRING = """\
-    %(filename)s = '''\\
+    %(filename)s = b'''\\
 %(fileData)s'''
 
 """
@@ -226,7 +226,7 @@ def __init_resources():
 """
 
     ADD_FILE_TO_MEMFS = """\
-    wx.MemoryFSHandler.AddFile('XRC/%(memoryPath)s/%(filename)s', %(filename)s)
+    wx.MemoryFSHandler.AddFile('XRC/%(memoryPath)s/%(filename)s', memoryview(%(filename)s))
 """
 
     LOAD_RES_MEMFS = """\


jeremyd2019:
I've just noticed that this is not working right on python 2.7.  I don't know if that matters anymore or not, but if so, it seems using bytearray() instead of memoryview() works on both python 2.7 and python3 (at the expense of an extra copy).

RobinD42:
> I don't know if that matters anymore or not

While I don't think we need to put a lots of effort into keeping 2.7 compatibility, but if it's something simple we may as well.

Metallicow:
Well I figure by the end of the year, most of the lingering deprecations and stuff will be taken care of for py2.7 - 3.7/8, and py27 will be about 6mos after end of life, so most folks will wrapup/archive their last versions for 2.7 and probably all go straight 3.5+ from that point forward. I think most things left are relatively small or int div warnings or stuff six will fix if you are running py27. After the end of the year, I won't put much effort into py27 anymore, so get those fixes in while people are still using it. My programs have been ready for the switch for at least a year now.

jeremyd2019:
> While I don't think we need to put a lots of effort into keeping 2.7 compatibility, but if it's something simple we may as well.

PR #1688 shows it is pretty simple.  I don't know much about sip or python memory buffers, but it seems a little odd in 
https://github.com/wxWidgets/Phoenix/blob/b452a353ed0c9e1aeb433c22b4abdac5206604c2/src/wxpybuffer.sip#L26-L29
it checks for both new and old buffer interface, but in 
https://github.com/wxWidgets/Phoenix/blob/b452a353ed0c9e1aeb433c22b4abdac5206604c2/src/wxpybuffer.h#L31-L44
it seems to only use the old interface on Python 2.  So there may be a more 'involved' fix for memorybuffer() on python 2.7.

RobinD42:
I think you may be misreading, or I'm misunderstanding your comment, but there is a `#if PY_MAJOR_VERSION < 3` in both cases.

PR #1680, opened 2020-06-15
Preserve `bytes` immutability.

mesalu:
`nanosvg` modifies its input `char` arrays in place (see: https://github.com/wxWidgets/nanosvg/blob/9dd92bbfc6f709e889578a724724ffbd42ac3004/src/nanosvg.h#L348),
which will lead to issues attempting to use the same python `bytes` object multiple times. Since bytes objects promise immutability its undesirable to permit them to be modified like this.
To correct for this, we'll copy the bytes object into a new location in memory and pass that new memory location to nanosvg.


RobinD42:

PR #1690, opened 2020-06-19
remove version from demo Main.py

Metallicow:
see https://github.com/wxWidgets/Phoenix/commit/88ff7b169721ad4ff7c8d9dd4813a9fe799ae420

now that file isn't generated anymore, this was breaking the demo framework from starting.

### DIFF
--- a/demo/Main.py
+++ b/demo/Main.py
@@ -72,8 +72,6 @@ from six import exec_, BytesIO
 from six.moves import cPickle
 from six.moves import urllib
 
-import version
-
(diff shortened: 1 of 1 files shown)


RobinD42:
But it is still generated when doing a build. The problem is that we're not ensuring that it is present when the source or demo tarballs are made.

RobinD42:
Superseded by 4ac15dda3a4ee237f4b4e34324695dadc817f286

Metallicow:
thanks

Issue #1526, opened 2020-02-25
Pure Python versions of widgets


Metallicow:
We really need pure-python versions of every widget added to the main lib. 
Too many users are asking for example a dropin checkbox or radiobutton/group that is customisable to some extent. etc...

- [ ] Checkbox
- [ ] Radiobutton
- [ ] RadioGroup
.... other stuff .... *post comments if you have other basic widgets that would apply here*

```python
# Implement Basic widgets in pure python

class CustomCheckbox(wx.Contol):
     .....
```

I'm sure some companies/coders already have already done this numerous times before, lets make it finally done as a boilerplate code.

infinity77:
One it’s already there, just needs to be converted to Phoenix syntax:

https://wiki.wxpython.org/CreatingCustomControls

Metallicow:
Ah, yea, I briefly recall going over that stuff at one time in the beginnings.  Kinda forgot the wiki is still useful even for old stuff. That should cover checkbox and most of radiobox functionality. Just needs converted to phoenix...

Metallicow:
Actually I think my RearrangeList python implementation applies too, since I was porting it to work with classic. Same idea, just make a python version of it...
https://github.com/Metallicow/MCOW/blob/master/wxPython_Phoenix/mcow/rearrangelist.py

Metallicow:
CheckListBox is another.....
See https://github.com/wxWidgets/Phoenix/issues/1692
Despite listing them all, They should come one as one.

PR #1695, opened 2020-06-22
Add PurePython Checkbox

Metallicow:
Mostly ported from the wxPyWiki impl and also is a mixin to get bitmaps if desired if hacking on other widgets.

Related to pure python impl issue #1526

Screenie win py2/3
purepy_checkbox

@RobinD42 @infinity77 Small sample app attached at bottom of file to test. Can I get a quick review nothing broke on your ends? Some of the comments probably need updated since classic and maybe some convienience methods added... but it works for me.


Metallicow:
I'm guessing this should be fine. I'm not the best documentor... :/
Moved to checkbox.py
Did best to match other files like buttons naming conventions.
appended _ to helper functions.
Sphinxified as best I could.
Made Demo

wx/lib/checkbox.py
```diff
@@ -0,0 +1,807 @@
...
+        :type `validator`: `wx.Validator`
+        :param `name`: Window name.
+        :type `name`: str
+        """
+
+        # Ok, let's see why we have used wx.PyControl instead of wx.Control.
+        # Basically, wx.PyControl is just like its wxWidgets counterparts
+        # except that it allows some of the more common C++ virtual method
+        # to be overridden in Python derived class. For GenCheckBox, we
+        # basically need to override DoGetBestSize and AcceptsFocusFromKeyboard.
```
RobinD42, line 104: Using wx.PyControl is no longer needed, so you can remove this comment about why you're using it.

wx/lib/checkbox.py
```diff
@@ -0,0 +1,807 @@
...
+        # Then we want to monitor user clicks, so that we can switch our
+        # state between checked and unchecked.
+        self.Bind(wx.EVT_LEFT_DOWN, self.OnMouseClick)
+        if wx.Platform == '__WXMSW__':
+            # MSW Sometimes does strange things...
+            self.Bind(wx.EVT_LEFT_DCLICK,  self.OnMouseClick)
+
+        # We want also to react to keyboard keys, namely the
+        # space bar that can toggle our checked state.
+        self.Bind(wx.EVT_KEY_UP, self.OnKeyUp)
```
RobinD42, line 146: I think checkboxes typically toggle on the key-down event.

wx/lib/checkbox.py
```diff
@@ -0,0 +1,807 @@
...
+        # while the height is the maximum value between the text width and
+        # the bitmap width.
+        totalWidth = bitmapWidth + spacing + textWidth
+        totalHeight = max(textHeight, bitmapHeight)
+
+        best = wx.Size(totalWidth, totalHeight)
+
+        # Cache the best size so it doesn't need to be calculated again,
+        # at least until some properties of the window change.
+        self.CacheBestSize(best)
```
RobinD42, line 275: This may be unnecessary. The caller of `DoGetBestSize` should already be passing the size to `CacheBestSize`

wx/lib/checkbox.py
```diff
@@ -0,0 +1,807 @@
...
+        """ Overridden base class virtual. """
+
+        # We can accept focus from keyboard, obviously.
+        return True
+
+    def AcceptsFocus(self):
+        """ Overridden base class virtual. """
+
+        # It seems to me that wx.CheckBox does not accept focus with mouse
+        # but please correct me if I am wrong!
```
RobinD42, line 289: It does. You just can't see the focus rectangle until there's another keypress or navigation event (at least on some platforms.)

wx/lib/checkbox.py
```diff
@@ -0,0 +1,807 @@
...
+        Handles the ``wx.EVT_ERASE_BACKGROUND`` event for :class:`GenCheckBox`.
+
+        :param `event`: A `wx.EraseEvent` to be processed.
+        :type `event`: `wx.EraseEvent`
+        """
+
+        # This is intentionally empty, because we are using the combination
+        # of wx.BufferedPaintDC + an empty OnEraseBackground event to
+        # reduce flicker.
+        pass
```
RobinD42, line 547: Since the paint event draws the whole widget, then you can use `SetBackgroundStyle(wx.BACKGROUND_STYLE_PAINT)` and then implementing an erase-background handler is not neccessary.

wx/lib/checkbox.py
```diff
@@ -0,0 +1,807 @@
...
+            "CheckedDisable": _GetCheckedImage(self).ConvertToDisabled().ConvertToBitmap(),
+            "UnCheckedDisable": _GetNotCheckedImage(self).ConvertToDisabled().ConvertToBitmap()}
+
+    def InitializeColours(self):
+        """ Initializes the focus indicator pen. """
+
+        textClr = self.GetForegroundColour()
+        if wx.Platform == "__WXMAC__":
+            self._focusIndPen = wx.Pen(textClr, 1, wx.PENSTYLE_SOLID)
+        else:
```
RobinD42, line 174: Pens on Mac support the dashed style now so this isn't needed.

wx/lib/checkbox.py
```diff
@@ -0,0 +1,807 @@
...
+        """
+        Handles the ``wx.EVT_PAINT`` event for :class:`GenCheckBox`.
+
+        :param `event`: A `wx.PaintEvent` to be processed.
+        :type `event`: `wx.PaintEvent`
+        """
+
+        # If you want to reduce flicker, a good starting point is to
+        # use wx.BufferedPaintDC .
+        dc = wx.BufferedPaintDC(self)
```
RobinD42, line 461: `wx.AutoBufferedPaintDC` would be marginally better

wx/lib/checkbox.py
```diff
@@ -0,0 +1,807 @@
...
+
+# -----------------------------------------------------------------------------
+
+
+class DefineNativeCheckBoxBitmapsMixin():
+    """
+    Inherit in your :class:`wx.Window` based subclass and call
+    ``self.DefineNativeCheckBoxBitmaps()`` in your __init__ def or startup routine
+    to define the native CheckBox Bitmaps as attributes the user can customize other
+    widgets appearance with.
```
RobinD42, line 621: This sentence is not easy to parse. Perhaps separate it into what-is-this-class and how-you-use-it, and maybe a simple example.

wx/lib/checkbox.py
```diff
@@ -0,0 +1,807 @@
...
+    """
+    def DefineNativeCheckBoxBitmaps(self):
+        """
+        Define native checkbox bitmaps as attributes. Returns True if all bitmaps
+         was defined Ok.
```
RobinD42, line 626: indentation

RobinD42 (changes requested):
It's looking good, just a few more minor things needed.

wx/lib/checkbox.py
Metallicow: removed

wx/lib/checkbox.py
Metallicow: on win7 the checkbox actually changes on up, tho while spacebar it is down it shows pressed graphic. Might all be OS specific but this is just intended to be a basic 2 clicker at the moment that the programmer can deal with if they want to mod it more and get more specific on the platform. os tweaks will be welcome somehow/method if others want to be completely anal about it and send a pull reqz.

wx/lib/checkbox.py
Metallicow: I'm not sure. by what you said, That basically means to me it is a idiot-proofing line. So you might want to check if what I understand your trying to say is that it may be being called 2twice...? Did not modify this one.

wx/lib/checkbox.py
Metallicow: I changed it to true. False was keeping keyboard focus, but changing to True makes the focus rect always draw. Could be improved if someone wants to be anal about the native appearance 100%. having a focus rect is better than it not working right from spacebar and activating wrong checkbox.

wx/lib/checkbox.py
Metallicow: Removed, but added in a OnSize refresh to prevent smearing.

wx/lib/checkbox.py
Metallicow: removed these lines

wx/lib/checkbox.py
Metallicow: Changed to Auto and added commenting

wx/lib/checkbox.py
Metallicow: Added basic gist of how I started MyCheckListBoxSTC as a sample example and made wording a bit clearer.
Also added current and pressed to the defined bitmaps.

wx/lib/checkbox.py
Metallicow: fixed

Metallicow:
Ok, you might want to rereview again and if it works right on mac, then it works good enough on windows for a basic pure py checkbox.

PR #1709, opened 2020-07-02
Try both old and new buffer protocols on python2.7

jeremyd2019:
This is what I was trying to say in https://github.com/wxWidgets/Phoenix/pull/1679#issuecomment-647053707

 I don't know if this is right or not, just trying to communicate via code rather than English is sometimes more effective.

### DIFF
--- a/src/wxpybuffer.h
+++ b/src/wxpybuffer.h
@@ -31,8 +31,9 @@ public:
         #if PY_MAJOR_VERSION < 3
             // Old buffer protocol
             int rv = PyObject_AsReadBuffer(obj, (const void**)&m_ptr, &m_len);
-            return rv != -1;
-        #else
+            if (rv != -1)
+                return true;
+        #endif
             // New buffer protocol
             Py_buffer view;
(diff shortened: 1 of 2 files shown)


RobinD42:
Oh, now I see what you were trying to say. Thanks for the code-speak. :wink:

It is correct to not use the old buffer protocol as it has been deprecated since Python 3.0.  See https://docs.python.org/3/c-api/objbuffer.html

jeremyd2019:
> It is correct to not use the old buffer protocol as it has been deprecated since Python 3.0. See https://docs.python.org/3/c-api/objbuffer.html

I thought what was going on was *only* the OLD protocol was being used on python < 3, while both old AND new were being checked for in the .sip file.  My change was to attempt to use both on python < 3 in the usage to match the check in the .sip file.

jeremyd2019:
I set up a build environment and was able to verify that this change does in fact make memoryview() work on python2.7.  I will clean this up (rebase the merge commit away and revert the switch to bytearray) and then switch it out of draft mode.

PR #1711, opened 2020-07-02
Add `IsNull()` checks for all property methods in GraphicsObject classes

### LINKED ISSUE #1706: Crash when accessing wx.NullGraphicsPath.Box
**Operating system**: Win10
**wxPython version & source**: pypi v4.1.0
**Python version & source**:   official `Python 3.8.0 (tags/v3.8.0:fa919fd, Oct 14 2019, 19:21:23) [MSC v.1916 32 bit (Intel)] on win32`

Same as #1479, but for `wx.NullGraphicsPath.Box`.

**Steps to reproduce**:
Run this code:
```python
import wx; wx.NullGraphicsPath.Box
```
-> crash

---
It might be a good idea to have a unittest like this, to catch similar errors in the future:
```python
for a in [getattr(wx, x, None) for x in dir(wx)]:
    for b in [getattr(a, x, None) for x in dir(a)]:
       for c in [getattr(b, x, None) for x in dir(b)]:
          pass
```
But that also triggers some assertions and opens popup boxes. I don't know if these are bugs or desired behavior, though.

RobinD42:
Fixes #1706

### DIFF
--- a/etg/graphics.py
+++ b/etg/graphics.py
@@ -201,11 +201,6 @@ def run():
     c.find('GetCurrentPoint').findOverload('wxDouble *x, wxDouble *y').ignore()
     c.mustHaveApp()
 
-    c.find('GetNativePath').setCppCode("""\
-        if (self->IsNull())
-            return (void*)0;
-        return self->GetNativePath();
-        """)
 
     #---------------------------------------------
     c = module.find('wxGraphicsRenderer')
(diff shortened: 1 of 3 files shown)


etg/graphics.py
```diff
@@ -336,6 +320,35 @@ def wrapper(self, obj):
 
     #-----------------------------------------------------------------
     tools.doCommonTweaks(module)
+
+    # Add some code to check obj.IsNull() to all methods that are used as getters for a
+    # PropertyDef. This is needed because it seems that most methods in GraphicsOpbects
+    # assume that the dev has already checked that the object is valid and so don't check
+    # it themselves. But when turned into a Python property they will automatically be called
+    # when introspecting the property values in things like wxNullGraphicsFOO. This cas
```
Chronial, line 328: > This *cas*

small typo

Chronial:
I don't know how easy this is to do, but when the methods are exposed as attributes, The `ValueError` should probably be rethrown as an `AttributeError`?

RobinD42:
Raising an exception is the only way I could protect things generically, and automatically apply the code to all cases. Otherwise a custom method for each instance would be required, which I like to avoid as much as possible.

PR #1749, opened 2020-07-27
Unittest fixes

StefanBruens:
Several fixes for the unittests

### DIFF
--- a/unittests/test_gdicmn.py
+++ b/unittests/test_gdicmn.py
@@ -15,7 +15,7 @@ class TestPoint(unittest.TestCase):
 
     def test_RealPoint_ctor(self):
         p = wx.Point(wx.RealPoint(1.2, 2.9))
-        self.assertTrue(p == (1,2))
+        self.assertTrue(p == (1,3))
 
     def test_copy_ctor(self):
         p1 = wx.Point(3,4)
@@ -299,8 +299,8 @@ class TestSize(unittest.TestCase):
 
         assert v1 == (12, 16)
         assert v2 == (12, 16)
-        assert v3 == (13, 17)
-        assert v4 == (14, 19)
+        assert v3 == (13, 18)
+        assert v4 == (15, 20)
 
 
 #---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


unittests/test_gdicmn.py
```diff
@@ -15,7 +18,10 @@ def test_xy_ctor(self):
 
     def test_RealPoint_ctor(self):
         p = wx.Point(wx.RealPoint(1.2, 2.9))
-        self.assertTrue(p == (1,2))
+        if viTuple < (3, 1, 2):
```
swt2c, line 21: Why support wx < 3.1.2?

StefanBruens: Sou you can have identical code in the 4.0 and 4.1 branches

swt2c: I don't think Robin is really maintaining the 4.0.x branch any more.  Certainly test fixes wouldn't be likely to be backported.

RobinD42: The master branch is for 4.1 and beyond, it would be undesirable to clutter it up for things that are specific to a prior release series like 4.0. Changes and PRs that are 4.0.x specific would need to be applied to the wxPy-4.0.x branch, although as @swt2c said, it's unlikely that there will be another release in that series. 

Also, it would have been more appropriate to use `wx.VERSION` anyway, instead of the wxWidgets version number.

StefanBruens: > Also, it would have been more appropriate to use `wx.VERSION` anyway, instead of the wxWidgets version number.

Definitely not - wx.VERSION >= 4.1 refers to _any_ wxWidgets version >= 3.1, but this changed between 3.1.1 and 3.1.2.

RobinD42: Oh, I see. I didn't look closely enough to see that's where you were making the demarcation. Regardless, `master` should be forward looking. If there's ever a need to make a build from wxWidgets 3.1.1 then we'll just checkout the wxPython revision that has a submodule link to the desired revision of wxWidgets, and end up with the older version of the test code as well.

PR #1759, opened 2020-08-04
Add agw.aui.AUI_DOCKART_HINT_WINDOW_BORDER_COLOUR theming

Metallicow:
Hint window border color can be set now with AUI_DOCKART_HINT_WINDOW_BORDER_COLOUR

Optimizations:
Optimized venetian blind loop and removed unnecessary wxPoint to tuple in Paint method wx.Rect

Screenie
AUI_DOCKART_HINT_WINDOW_BORDER_COLOUR

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -58,6 +58,9 @@ New and improved in this release:
 * Simplified the implementation of the wx.App.InitLocale method. See the
   MigrationGuide for more information.
 
+* Added wx.lib.agw.aui.AUI_DOCKART_HINT_WINDOW_BORDER_COLOUR constant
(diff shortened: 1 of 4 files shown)


Metallicow:
@infinity77  @RobinD42 ... would it be helpful also to add a AUI_DOCKART_HINT_WINDOW_BORDER_WIDTH constant for the hint window...? ... or would scaling on high rez monitors fix/double the border width...

Also the Hint window could be used outside of agw.aui and probably would be better off ripped out and added to wx.lib.
Thoughts?
... Tho ripping the docking guides out to be their own widget would be a bit more of an issue/sideproject as they are pretty well grained into the agw.aui code...

PR #1775, opened 2020-08-22
Do not mangle type of wxART_* identifiers

### LINKED ISSUE #1768: Demo ArtProvider failed
**wxPython version & source**: Newest from Github        
**Python version & source**: Python 3.8           

If I run the wxPython demo, the `ArtProvider` failed with an error. See snapshot.
The error is raising since version `wxPython-4.1.1a1.dev4919+57702fa7`. In earlier versions all works well.

pythonw_2020-08-12_11-52-43

StefanBruens:
Starting with wxWidgets 3.1.4, the identifiers are no longer const char[N].

Unfortunately, the implementation in 3.1.4 does not provide wxStrings
with static storage duration but just temporaries, so this can not be
used as constants from Phoenix, though 3.1.5 fixes this.

Fixes #1768 
See wxWidgets/wxWidgets#1996

### DIFF
--- a/etg/artprov.py
+++ b/etg/artprov.py
@@ -52,13 +52,10 @@ def run():
     # deprecated and removed
     c.find('Insert').ignore()
 
-    # Change the types of the art constants from wxString to const char*
-    # since that is what they really are.
     artConsts = list()
     for item in module:
         if isinstance(item, etgtools.GlobalVarDef):
             if item.type in ['wxArtClient', 'wxArtID']:
-                item.type = 'const char*'
                 artConsts.append(item)
     # move them to the front of the module
     for item in artConsts:


RobinD42:
I'm going to push a temporary workaround so there can be a snapshot or two made before updating to the newest wxWIdgets master. I'll probably to the update and make this change or a similar one later this week.

RobinD42:
Temporary workaround is in 71c7514593567435feb8bc5ad93f097c8ac94fc1

RobinD42:
Or maybe not. The relevant wxWidgets change was reverted today and things were moved back to closer to what they were originally. But it looks like I can still improve what's in `artprov.py` for it.

RobinD42:
Superseded by #1793

PR #1807, opened 2020-10-01
Fix GLCanvas Demo

Metallicow:
Add a bit more example to the cube and the cone. Made note to DO NOT USE glut. It is old.
Folks seem to always have hard time figuring out how to load images. Added image texturing example.
Also changed cone to meshmode.

Fixes Broken GLCanvas.py Demo once again.

### DIFF
--- a/demo/GLCanvas.py
+++ b/demo/GLCanvas.py
@@ -1,7 +1,10 @@
 #!/usr/bin/env python
 
-import wx
+import os
 import sys
+from math import pi, sin, cos
+
+import wx
 
 try:
     from wx import glcanvas
@@ -13,11 +16,17 @@ try:
     # The Python OpenGL package can be found at
     # http://PyOpenGL.sourceforge.net/
     from OpenGL.GL import *
-    from OpenGL.GLUT import *
+    from OpenGL.GLU import *
+    #### OpenGL.GLUT  Very old, DO NOT USE!
     haveOpenGL = True
 except ImportError:
     haveOpenGL = False
 
+## try:
+##     from PIL import Image
+##     havePIL = True
+## except ImportError:
+##     havePIL = False
 #----------------------------------------------------------------------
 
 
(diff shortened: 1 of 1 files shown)


demo/GLCanvas.py
```diff
@@ -13,11 +16,17 @@
...
+    #### OpenGL.GLUT  Very old, DO NOT USE!
     haveOpenGL = True
 except ImportError:
     haveOpenGL = False
 
+## try:
+##     from PIL import Image
+##     havePIL = True
+## except ImportError:
+##     havePIL = False
```
RobinD42, line 29: remove this block

demo/GLCanvas.py
```diff
@@ -13,11 +16,17 @@
     # The Python OpenGL package can be found at
     # http://PyOpenGL.sourceforge.net/
     from OpenGL.GL import *
-    from OpenGL.GLUT import *
+    from OpenGL.GLU import *
+    #### OpenGL.GLUT  Very old, DO NOT USE!
```
RobinD42, line 20: Just remove the line, no need to preach.

demo/GLCanvas.py
```diff
@@ -62,14 +71,15 @@ def OnButton(self, evt):
             dlg = wx.MessageDialog(self,
                                    'The OpenGL package was not found.  You can get it at\n'
                                    'http://PyOpenGL.sourceforge.net/',
+                                   '/nor $ pip install PyOpenGL PyOpenGL_accelerate',
```
RobinD42, line 74:
```
  File "./GLCanvas.py", line 71, in OnButton
    dlg = wx.MessageDialog(self,
TypeError: MessageDialog(): argument 4 has unexpected type 'str'
```

RobinD42 (changes requested):

demo/GLCanvas.py
Metallicow: Opps. that one wasn't intended. Guess I had a brain fart. Fixed comma typo and formatting.
Also removed other requested lines.

PR #1817, opened 2020-10-12
Python 3.9

RobinD42:
Initial work for adding Python 3.9 builds


cclauss:
https://github.com/wxWidgets/Phoenix/blob/master/setup.py#L84-L88

Py34 and Py35 are EOL https://devguide.python.org/devcycle/#end-of-life-branches
Py38 and Py39 pass the CI tests https://devguide.python.org/#status-of-python-branches

PR #1826, opened 2020-10-19
Reduce flicker in svg demos

Metallicow:
Reduce flicker in svg demos

### DIFF
--- a/demo/SVGImage_Bitmap.py
+++ b/demo/SVGImage_Bitmap.py
@@ -21,6 +21,8 @@ class SVGBitmapDisplay(wx.Panel):
         sbox = wx.StaticBoxSizer(wx.VERTICAL, self, label)
         sbox.Add(self.statbmp)
         self.SetSizer(sbox)
+        if not self.IsDoubleBuffered():
+            self.SetDoubleBuffered(True)  # Reduce flicker on size event.
 
 
     def UpdateSVG(self, svg_filename):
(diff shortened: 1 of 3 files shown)


RobinD42:
1. I don't see any flicker in these two samples at all. Does anybody else? (On Windows which is really the only platform that needs to worry about flicker.)

2. Setting the panel to be double-buffered is not guaranteed to have any effect on its static bitmap children. Besides the static bitmap is a native widget so it is probably already as optimized as it can be.

3. The change in `wx/lib/throbber.py` does not belong in this PR.

Metallicow:
Re:
1. Yes it is on Windows that I normally see this happen. Sometimes it is a small intermittent flicker instead of all out flicker on each size event. Using BufferedDC usually fixes this.
2. Sometimes buffered doesn't fix the small amount of intermittent flicker and when that happens, it seems that setting the parent to be double buffered(True) does. But yea, I think that is a bit ugly way to put it.

The small intermittent flicker can be seen when resizing it a bit on windows. You may have to play around with the gripper resizing for a second or two to see it flicker, but it does flicker occasionally like once or twice and is just enough to be noticeable.
@RobinD42  Would this be better if I do a `__MSW__` Platform check? That would add an extra line the paint event one that would drop performance a bit. Maybe it would be better to add that comment there also...

3. Can do. will separate/revert throbber commit.

PR #1837, opened 2020-10-31
Avoid releasing the GIL in wxCustomDataObject.GetData

### LINKED ISSUE #1833: wxCustomDataObject.GetData shouldn't release the GIL
As discovered by @opoplawski in https://github.com/enthought/pyface/issues/784, the `wxCustomDataObject.GetData` method crashes (sometimes?) because it calls `PyMemoryView_FromBuffer` after releasing the GIL.  `GetData` shouldn't release the GIL.

I'll send a PR shortly to fix this.

swt2c:
This avoids a crash because GetData calls i_wxPyMakeBuffer which calls
Python C API, for which the GIL needs to be held.

Fixes #1833.

### DIFF
--- a/etg/dataobj.py
+++ b/etg/dataobj.py
@@ -335,7 +335,7 @@ def run():
     c.find('TakeData').ignore()
 
     c.find('GetData').ignore()
-    c.addCppMethod('PyObject*', 'GetData', '()', isConst=True,
+    c.addCppMethod('PyObject*', 'GetData', '()', isConst=True, pyHoldGIL=True,
         doc="Returns a reference to the data buffer.",
         body="return wxPyMakeBuffer(self->GetData(), self->GetSize());")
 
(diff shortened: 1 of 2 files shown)


RobinD42:
The GIL should be reacquired inside of `i_wxPyMakeBuffer` with the creation of the `wxPyThreadBlocker`. (It acquires the GIL if it doesn't already have it.) That line was added in master about 6 months ago in 6885cf4226acbb524012f600affd0a450cce525b. I'd rather see that change backported to your 4.0.7 builds than to apply this change.

Metallicow:
I haven't noticed a crash in SourceCoder that involves this(tho I haven't specifically tested for it), so you may take Robin's advice into consideration before reconsidering.......

swt2c:
Agreed, sorry for not realizing this was already fixed in master.

PR #1866, opened 2020-12-05
Update Fedora docker builds

swt2c:
Remove Fedora 29-31 as they are now EOL.  Add Fedora 32 & 33.


swt2c:
Also, can you enable Fedora 33 for snapshot builds for my sake?  :)

RobinD42:
I'm adding these new DockerFiles in a different commit. (See c422227c1d9c62fd71d6cf8c2081dc7b2dcb7519). I've been keeping the old ones around to make it easy to use or rebuild an old image again if needed, but adding them to a list in `tasks.py` so they are not selected by default.

I have some changes in-progress to allow running (at least some of) the docker-based builds via the buildbot server where the snapshots and release builds are done. I'll try to get that finished up soonish.

PR #1895, opened 2021-01-11
Fixes transparency in the docking guides for wx.lib.agw.aui

### LINKED ISSUE #1894: Docking guides in wx.lib.agw.aui do not have the frame shaped correctly
wx.Region no longer selects all pixels that have an alpha channel other then 255 when a wxBitmap is passed to the constructor. the color that contains the alpha channel that the region is supposed to have subtracted from the bitmap has to be supplied.

kdschlosser:
wx.Region no longer selects all non transparent pixels when a wxBitmap is passed to the constructor. In order to get the region made properly a color bust also be supplied. As luck would have it the way the docking guide images were made we are able to do this because there is only a single color that we needed to use.

Fixes #1894

### DIFF
--- a/wx/lib/agw/aui/framemanager.py
+++ b/wx/lib/agw/aui/framemanager.py
@@ -2298,7 +2298,7 @@ class AuiSingleDockingGuide(AuiDockingGuide):
 
         useAero = (useWhidbey and [2] or [1])[0]
         bmp, dummy = GetDockingImage(self._direction, useAero, False)
-        region = wx.Region(bmp)
+        region = wx.Region(bmp, wx.Colour(0, 0, 0, 0))
 
         self.region = region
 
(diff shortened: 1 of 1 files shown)


RobinD42:
Have you tried a recent snapshot build? I'm pretty sure that the mask issues have been fixed already, at least if the source image actually has a mask and not an alpha channel. The root of the change was that images with alpha don't automatically get a mask anymore.

kdschlosser:
@RobinD42 

I am not sure if the images have a mask or not, they are embedded into a python file. If I had to guess they probably do not have a mask and there is only the alpha channel. If memory serves wxRegion used to create the region so it omitted any pixel that had an alpha of < 255 when a wxBitmap instance was passed to it. maybe I am wrong in this it has been a while since I messed about with regions.

I know by adding the color (0, 0, 0, 0) to the call to the constructor of wxRegion the problem is solved.

kdschlosser:
Thanks Robin. :+1:

PR #1896, opened 2021-01-11
Fixes floating point math in wx.lib.agw.aui

### LINKED ISSUE #1881: Application crash using AGW AUI
**Operating system**: Windows 7 x64
**wxPython version & source**: wxPython 4.1.1 (pypi)
**Python version & source**:    CPython 3.8.6 (python.org)

When docking a pane I get an application crash.
0xC0000005: Access violation

The crash occurs in wxbase315u_vc140_x64.dll
Line 1637 of wxWidgets/src/common/event.cpp `return ProcessEvent(event);`
in function `wxEvtHandler::SafelyProcessEvent(wxEvent & event)`

the `wx.Frame` has been constructed with these flags
wx.MINIMIZE_BOX
wx.MAXIMIZE_BOX
wx.RESIZE_BORDER
wx.CAPTION
wx.SYSTEM_MENU
wx.CLOSE_BOX
wx.CLIP_CHILDREN
wx.TAB_TRAVERSAL

the `wx.lib.agw.aui.AuiManager` has been constructed with these flags
aui.AUI_MGR_ALLOW_FLOATING
aui.AUI_MGR_ALLOW_ACTIVE_PANE
aui.AUI_MGR_TRANSPARENT_DRAG
aui.AUI_MGR_VENETIAN_BLINDS_HINT
aui.AUI_MGR_HINT_FADE
aui.AUI_MGR_LIVE_RESIZE
aui.AUI_MGR_ANIMATE_FRAMES
aui.AUI_MGR_WHIDBEY_DOCKING_GUIDES
aui.AUI_MGR_PREVIEW_MINIMIZED_PANES
aui.AUI_MGR_SMOOTH_DOCKING

the `wx.lib.agw.aui.AuiPaneInfo` as been constructed using these options
Name(name)
Caption(caption)
Row(0)
MinSize((300, 300))
FloatingSize((300, 300))
MinimizeMode(aui.AUI_MINIMIZE_POS_BOTTOM | aui.AUI_MINIMIZE_CAPT_SMART)
Resizable(True)
Show()
CaptionVisible(True, left=True)
PaneBorder(True)
Gripper(True)
GripperTop(True)
CloseButton(True)
MaximizeButton(True)
MinimizeButton(True)
PinButton(True)
DestroyOnClose(True)
Floatable(True)
Movable(True)
Dockable(True)
Snappable(True)
FlyOut(True)

the "window" the pane is being populated with is a `wx.ListCtrl` instance

What I am doing is I am clicking on the gripper of the pane and moving it, once I move it the docking guides appear. I move the mouse over the right arrow in the docking guide. It then shows where the pane will be docked, I then let go of the mouse button the pane gets docked and then application crash.

I am not binding to any AUI events so I know I am not inadvertently stepping on somethings toes.

I am sub classing `wx.lib.agw.aui.framemanager.AuiPaneInfo` only for the purposes of setting up the pane and the only method I have overridden is \_\_init\_\_. Could this cause the problem?

I have attached the Visual Studio debug dump file.
[agw_aui_crash_dump.zip](https://github.com/wxWidgets/Phoenix/files/5728096/agw_aui_crash_dump.zip)

Thanks for the help

### LINKED ISSUE #1885: AGW AUI crash on detaching pane with Python 3.9.1
**Operating system**: Windows 10 x64
**wxPython version & source**: wxPython 4.1.1 (PyPI wheel)
**Python version & source**: CPython 3.9.1 32 bit (python.org)

I experience a crash when running AGW AUI sample code from official documentation: https://wxpython.org/Phoenix/docs/html/wx.lib.agw.aui.html

When I drag "Pane Number One" or "Pane Number Two", as soon as I detach the pane from the window side, my application crashes without any error messages.

It works fine in Python 3.8.7.

kdschlosser:
wx.lib.agw.aui was never converted to run on Python 3 and as a result the division operator forces a conversion of the used values to a float. Floats cannot be used in functions like `range`. This caused quite number of things to not function properly and produce tracebacks.

The other thing is when wx.lib.agw.aui was written pixels could not be rendered using coordinates that were floats so now passing floats could cause rendering alignment problems because the layout of the various bits were done so in a manner that would have the alignment correct using integers.

Fixes #1881
Fixes #1885

This does not fix the application crash that is caused by setting AuiPaneInfo.FlyOut to `True`. That problem is caused by something internal to wxTimer

### DIFF
--- a/wx/lib/agw/aui/aui_switcherdialog.py
+++ b/wx/lib/agw/aui/aui_switcherdialog.py
@@ -621,12 +621,12 @@ class SwitcherItems(object):
                    and item.GetBitmap().GetHeight() <= 16:
                     x -= textMarginX
                     dc.DrawBitmap(item.GetBitmap(), x, item.GetRect().y + \
-                                  (item.GetRect().height - item.GetBitmap().GetHeight())/2,
+                                  (item.GetRect().height - item.GetBitmap().GetHeight())//2,
                                   True)
                     x += 16 + textMarginX
                 #x += textMarginX
 
-            y = item.GetRect().y + (item.GetRect().height - h)/2
+            y = item.GetRect().y + (item.GetRect().height - h)//2
             dc.DrawText(item.GetTitle(), x, y)
             dc.DestroyClippingRegion()
 
--- a/wx/lib/agw/aui/aui_utilities.py
+++ b/wx/lib/agw/aui/aui_utilities.py
@@ -322,9 +322,9 @@ def DarkenBitmap(bmp, caption_colour, new_colour):
     """
 
     image = bmp.ConvertToImage()
-    red = caption_colour.Red()/float(new_colour.Red())
-    green = caption_colour.Green()/float(new_colour.Green())
-    blue = caption_colour.Blue()/float(new_colour.Blue())
+    red = caption_colour.Red()/new_colour.Red()
+    green = caption_colour.Green()/new_colour.Green()
+    blue = caption_colour.Blue()/new_colour.Blue()
     image = image.AdjustChannels(red, green, blue)
     return image.ConvertToBitmap()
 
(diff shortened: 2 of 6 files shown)


kdschlosser:
Thanks Robin. :+1:

PR #1900, opened 2021-01-14
Update hypertreelist.py

### LINKED ISSUE #1899: HyperTreeList should no longer set wx.ListEvent m_ properties
**Operating system**:                        Windows 10
**wxPython version & source**:         4.1.0 and 4.1.1 PyPI
**Python version & source**:             3.8.3 installer downloaded from Python website

**Description**: 
The Phoenix migation guide has the following about wx.ListCtrl:
“In wx.ListItem and wx.ListEvent the “m_” properties are no longer public. Instead use the associated getter/setter methods or the auto-generated properties that are using them.”

When sending a wx.ListEvent though (for EVT_COMMAND_LIST_COL_DRAGGING, EVT_COMMAND_LIST_COL_BEGIN_DRAG, EVT_COMMAND_LIST_COL_END_DRAG or EVT_COMMAND_LIST_COL_CLICK), wx.lib.agw.hypertreelist.HyperTreeList still sets m_pointDrag and m_col. Presumably these settings should be replaced with calls to evt.SetColumn(…) and evt.SetPoint(…)? This causes a problem where evt.GetColumn() does not return the correct event column index (for me it returns 0 or -1), while using evt.m_col does.

To test with the code below for a small sample app, drag different column headers to resize the columns; the values of event.GetColumn() and event.m_col will be reported.

```
#!/usr/bin/env python
""" Test app """
import wx
import wx.lib.agw.hypertreelist as HTL

class MainFrame(wx.Frame):
    def __init__(self):

        wx.Frame.__init__(self, None, wx.ID_ANY)

        # Contains a HyperTreeList
        tree = HTL.HyperTreeList(self, wx.ID_ANY)

        for icol in range(4):
            tree.AddColumn(f"Column {icol}")
        tree.SetMainColumn(0)

        root = tree.AddRoot("The root item")

        tree.Bind(wx.EVT_LIST_COL_END_DRAG, self.OnColResize)


    def OnColResize(self, evt):
        try:
            print("event.GetColumn() = ", evt.GetColumn())
        except AttributeError:
            print("event has no GetColumn method")

        try:
            print("event.m_col = ", evt.m_col)
        except AttributeError:
            print("event has no m_col property")
        print("")


# Application loop
print(wx.version())
app = wx.App(False)
mainwin = MainFrame()
mainwin.Show(True)
app.MainLoop()

```

JillHolliday:
Replaced setting of wx.ListEvent m_ properties with calls to setter methods

Replaced the SendListEvent method code that was setting the event m_col and m_pointDrag properties with calls to SetColumn(...) and SetPoint(...) respectively.

Note that user code still accessing m_col or m_pointDrag will need changes (although if this is undesirable a further change could be made to set these properties too).

Fixes #1899

### DIFF
--- a/wx/lib/agw/hypertreelist.py
+++ b/wx/lib/agw/hypertreelist.py
@@ -1136,14 +1136,13 @@ class TreeListHeaderWindow(wx.Window):
         parent = self.GetParent()
         le = wx.ListEvent(evtType, parent.GetId())
         le.SetEventObject(parent)
-        le.m_pointDrag = pos
 
         # the position should be relative to the parent window, not
         # this one for compatibility with MSW and common sense: the
         # user code doesn't know anything at all about this header
         # window, so why should it get positions relative to it?
-        le.m_pointDrag.y -= self.GetSize().y
-        le.m_col = self._column
+        le.SetPoint((pos.x, pos.y - self.GetSize().y))
+        le.SetColumn(self._column)
         parent.GetEventHandler().ProcessEvent(le)
 
 


JillHolliday:
Apologies, I thought I'd checked thoroughly that this hadn't already been fixed, but having submitted this I just spotted a comment in PR #1128 which makes me think it's already been addressed.

JillHolliday:
I'm closing this PR as it's covered by PR #1128 (and doesn't pass all the checks either)

PR #1916, opened 2021-01-31
Add support for building with SIP v6

mitya57:
SIP v6 [was released this month][1], and it is what users now get when installing [`sip` package][2] from PyPI.

In SIP v6, `sipName` argument was removed from `generateCode` function and added to `set_globals`. See [this changeset][3].

`SIP_VERSION` is a number like 0xABCDE, where 0xA is the major version, 0xBC — minor version, 0xDE — patch version (so for SIP 6.0.0 it is 0x60000, for sip 5.5.0 it is 0x50500). So by calling `SIP_VERSION >> 16` we get the major version.

[1]: https://riverbankcomputing.com/news/SIP_v6.0.0_Released
[2]: https://pypi.org/project/sip/
[3]: https://riverbankcomputing.com/hg/sip/rev/a16b972e11a1

### DIFF
--- a/buildtools/wxpysip.py
+++ b/buildtools/wxpysip.py
@@ -62,8 +62,12 @@ def sip_runner(
         abi_major, abi_minor = resolve_abi_version(abi_version).split('.')
 
         # Set the globals.
-        set_globals(SIP_VERSION, SIP_VERSION_STR, int(abi_major), int(abi_minor),
-                UserException, include_dirs)
+        if (SIP_VERSION >> 16) >= 6:
+            set_globals(SIP_VERSION, SIP_VERSION_STR, int(abi_major), int(abi_minor),
+                    sip_module, UserException, include_dirs)
+        else:
+            set_globals(SIP_VERSION, SIP_VERSION_STR, int(abi_major), int(abi_minor),
+                    UserException, include_dirs)
 
         # Parse the input file.
         pt, _, _, _, tags, disabled_features = parse(specification,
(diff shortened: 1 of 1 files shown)


swt2c:
Somewhat OT here, but do you plan to try to put sip6 in Bullseye?

mitya57:
I would like to do it, but don't know yet if I manage to do it in time before freeze.

wxpython4.0 and veusz are the remaining blockers, but veusz needs much more changes (it relied on legacy sip5 tool).

In any case, it would be nice if wxpython4.0 was made compatible with both sip5 and sip6. I was planning to create a merge request for packaging once this upstream PR is merged. If you want, I can do it earlier. Apart from this change, it needs build-dependency change from sip5-tools to sip-tools and a backport of one older upstream commit.

mitya57:
For those who want to test this with SIP v6: the current version of SIP contains a bug that causes UnicodeDecodeError when importing `wx` package. Upstream commit https://riverbankcomputing.com/hg/sip/rev/64d8a5d785e1 should fix it.

swt2c:
@RobinD42 what do you think about moving to sip 6, as sip 5 is EOL.  Presumably you would just want to move to sip 6 wholesale, rather than support both 5 and 6?

mitya57:
No, I have not looked and I did not realize that SIP 6.6 breaks API.

@swt2c says in #2169 that he is going to care of it.

jmoraleda:
@RobinD42 I am wondering if this can now be closed as completed by #2169.

swt2c:
wxPython 4.2.0 has been released and supports building with sip 6.6.2+

PR #1857, opened 2020-11-26
AuiToolBar checks that it has not been Destroyed by the most recent a…

jmoraleda:
`AuiToolBar` checks that it has not been Destroyed by the most recent action before doing any post-processing for the action. 
Fixes #1856 following the suggestion given there.

PR #1931, opened 2021-03-07
Fix regression in auibar introduced in last commit

jmoraleda:
This PR is identical to PR #1857 (merged in commit e059ec1a). The reason we need his PR again is that the bug fix in PR #1857 was inadvertently undone by commit  e32ba711.


jmoraleda:
Hi @RobinD42 What do you think?

PR #1941, opened 2021-03-28
Add configuration to use podman when building on Fedora

HelioGuilherme66:
When building wxPython from Fedora >=31 with docker, we see errors mentioning cgroup v2 incompatability. A possible solution is to install `podman` which replaces `docker`.
The changes in this PR allows to use `podman` and in rootless mode.

Tested on Fedora 32 to build for Python 3.9, with `fedora-33` image.


RobinD42:

https://discuss.wxpython.org/t/building-wxpython-on-fedora-32-with-docker/35277/1

PR #1993, opened 2021-08-05
README.rst: Encourage the use of current Python

cclauss:
Python 3.7 --> 3.9.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -361,7 +361,7 @@ or other unixes.
 
 * dpkg-dev
 * build-essential
-* python3.7-dev and libpython3.7-dev  # use appropriate Python version here
+* python3.9-dev and libpython3.9-dev  # use appropriate Python version here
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -361,7 +361,7 @@ or other unixes.
 
 * dpkg-dev
 * build-essential
-* python3.7-dev and libpython3.7-dev  # use appropriate Python version here
+* python3.9-dev and libpython3.9-dev  # use appropriate Python version here
```
swt2c, line 364: Might be even better just to specify `python3-dev` instead?  That will always pull in the current version.  Also, I think libpythonX.Y-dev is unnecessary.  It will get pulled in by python3-dev.

RobinD42:
Changed a little differently in 0cbf3d738a97057e5254756c9e8d8f5c98d4bb2a

PR #1998, opened 2021-08-07
unicode() is removed in Python 3

cclauss:
Fixes [`flake8 . --count --exclude=./buildtools/backports/six.py,./wx/lib/pubsub/py2and3.py --select=F82 --show-source --statistics | grep unicode`](https://pypi.org/project/flake8/)

### DIFF
--- a/build.py
+++ b/build.py
@@ -1065,7 +1065,7 @@ def _removeSidebar(path):
         tag = soup.find('div', 'document')
         if tag:
             tag.attrs['class'] = ['document-no-sidebar']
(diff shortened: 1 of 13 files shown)


RobinD42:
There are still some folks wanting Py 2.7 compatibility to continue for a while longer. I think that that ship has already sailed before this change, but I think I'll hold off merging it until they give up all hope.

cclauss:
Python 2 died on 1/1/2020.  That is 700+ days ago.

RobinD42:
I know. Some people believe in aliens too.

PR #2002, opened 2021-08-18
Fix functions of the shell

komoto48g:
This PR fixes some functionality of the wx.py.shell.

There was a problem that the Shell.OnUpdateUI overrides the parent EditWindow.OnUpdateUI. Thus the brace-matching is not working. In addition, the wx.EVT_UPDATE_UI is bound to the shell without specifying id, so the event is frequently triggered each time the caret moves or blinks. To fix it, we should remove the handler `OnUpdateUI` and bind wx.EVT_UPDATE_UI to each context-menu id.

It also fixes tiny bugs of incorrect indentation when pasting "try-finally" code and when typing "with" statement.


RobinD42:
Is there a reason `OnUpdateContextMenu` is only commented-out and not removed?

komoto48g:
Hi, Robin
Thank you for the code reviwing. There is no special reason.
(I'm sorry for including plural changes in this PR)
Best regards

RobinD42:
Okay. Please clean it up with an additional commit to your branch and then I'll merge it.

komoto48g:
I pushed them on the same PR branch.

PR #2005, opened 2021-08-31
fix: Correct types of arguments to ScrolledWindow.SetScrollbars

gwhitney:
The third and fourth arguments to ScrolledWindow.SetScrollbars are of integer
types and therefore need to be computed with expressions that result in an
integer. This PR removes the remaining places in the code where
ScrolledWindow.SetScrollbars was called with the float result of ordinary division,
eliminating warnings that the library was generating. The library should not
generate warnings in ordinary operation.

Thanks for reviewing and merging.


swt2c:
Thanks!

PR #2014, opened 2021-10-05
Add wxDC.DrawLinesFromBuffer to draw directly from numpy array memory using the Buffer protocol

arjones6:
When wxDC.DrawLines is called, it takes the list passed to it and creates a new wx.PointList, looping through the list and converting each element to a wx.Point.  For numpy arrays with the int32 type this adds unnecessary overhead as the memory is already structured as pairs of 32 bit integers and we can just cast the underlying memory buffer to the wxPoint * type and call DrawLines directly.  Doing this results in significant speed increases (>10x) particularly when there are a lot of points involved.

This pull request adds the DrawLinesFromBuffer function to the wx.DC class.  This function would use the python buffer protocol to access the underlying memory of an object and call DrawLines on it directly.  It checks to make sure each element is the right size, but doesn't check that it is the correct type.  An example usage would be:

        xs = numpy.linspace(0, 500, 10000)
        ys = 150 + numpy.sin(xs / 20) * 150
        xs.shape = xs.size, 1
        ys.shape = ys.size, 1
        pts = numpy.append(xs, ys, 1).astype('int32')
        dc.DrawLinesFromBuffer(pts)

Another approach might be to modify DrawLines to check if the passed object supports the buffer protocol and do it that way, but that could break currently functional code if somebody is passing something that supports the buffer protocol to DrawLines that isn't the correct type.

### DIFF
--- a/demo/DrawLinesFromBuffer.py
+++ b/demo/DrawLinesFromBuffer.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python
+
+import time
+
+import wx
+
+npts = 100000
+
+def TestLinesFromBuffer(dc, log):
+    global npts
+    try:
+        import numpy as np
+        start = time.time()
+
+        w, h = dc.GetSize()
+
+        vs = np.linspace(0, 1, npts).reshape((npts, 1))
+
+        x1 = np.cos(vs * 12 * np.pi) * w/2 * vs + w/2
+        x2 = np.cos(vs * 16 * np.pi) * w/2 * vs + w/2
+        y1 = np.sin(vs * 12 * np.pi) * w/2 * vs + h/2
+        y2 = np.sin(vs * 16 * np.pi) * w/2 * vs + h/2
+        
+
+        pts1 = np.append(x1, y1, 1).astype('int32')
+        pts2 = np.append(x2, y2, 1).astype('int32')
+
+        dc.SetPen(wx.BLACK_PEN)
+        t1 = time.time()
+        dc.DrawLines(pts1)
+        t2 = time.time()
+        
+        dc.SetPen(wx.RED_PEN)
+        t3 = time.time()
+        dc.DrawLinesFromBuffer(pts2)
+        t4 = time.time()
+        
+        log.write("%s pts: %s seconds with DrawLines %s seconds with DrawLinesFromBuffer\n" % (npts, t2 - t1, t4 - t3))
+
+    except ImportError:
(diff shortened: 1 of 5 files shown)


RobinD42:
Please add an example using this method somewhere in the demo. It doesn't need to be its own demo, it could make sense in the `DrawXXXList` sample. OTOH, if it was it's own demo you could easily show both `DrawLines` and `DrawLinesFromBuffer` drawing the same set of random points, and the time it took each to do their job.

Also, there should be a unittest added for it (which can be skipped if numpy isn't installed.)

Thanks!

arjones6:
I've added an example as a separate demo since it doesn't really fit in the DrawXXXList title, and made it so the user can change the number of points drawn.  If you want to combine them to keep the number of demos down, maybe renaming it to DC_Optimizations or something would make sense.

I added a unittest as well.  I'm not sure how consistent the wx.Point size is across different systems, so the test will fail if there are places where int32 isn't the appropriate size to match the underlying storage for a wx.Point.  If that will be an issue, is there some way to determine what the appropriate size should be?

Also, incorporating this into wx.lib.plot would provide a meaningful performance boost.  I held off on making any changes there since it would basically be the same as PR #1964 with DrawLinesFromBuffer on line 481 of polyobjects.  

Thanks!

RobinD42:
Yes, the size of the `wxPoint` elements could be an issue. The type used for `x` and `y` members are plain C integers (see https://github.com/wxWidgets/wxWidgets/blob/master/include/wx/gdicmn.h) and the size of `int` is platform dependent.

Numpy provides a dtype that is supposed to always match the C `int` type, `numpy.intc`. Since this function is supporting a generic buffer and not just numpy arrays there probably isn't anything that can be done specifically for numpy like verifying a size match with `numpy.intc`, but it could be used in the sample code with comments indicating why it's important to use it for compatibility with the C int. There should be a statement in the docstring about it too, that the elements in the buffer need to be C integers.

Thanks.

arjones6:
Ok.  I've updated the demo and test to use the intc data type and updated the documentation and demo explanation.  I do check that the buffer element size is the same as a wxPoint, so it will raise a TypeError if you use the wrong size integers.  There isn't any way to check (that I know of) what data type a buffer has though, so it doesn't complain if you use it with float data that matches the element size.  I haven't had it do anything other than not draw were it was supposed to though, so that doesn't seem like a big issue.

Let me know if there are any other changes you would like.  Thanks!

PR #2026, opened 2021-11-16
Fixed integer division in Python 3

carolfs:
This fixes integer division using the // operator instead of /

This bug was breaking fsleyes on Python 3.10/Fedora 35 for me.


swt2c:
Thanks!

PR #2029, opened 2021-11-18
Fix unbinding of events in FIFO order

### LINKED ISSUE #2027: Reversing unbind order breaks the event chain link and makes the program crash.
**Operating system**: Windows 10 64bit ver 20H2
**wxPython version & source**: 4.1.1 (pypi)
**Python version & source**: 3.8.6

In the sample code, the button object only binds and unbinds the two handlers. If the unbinding order is reversed, the program will crash. I also checked it with wx.version 4.0.7, and there was no problem.

<details>

```python
import wx

class TestPanel(wx.Panel):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        
        self.btn = wx.Button(self, label="Hello, wxPython!")
        
        self.btn.Bind(wx.EVT_BUTTON, self.onButton1)
        self.btn.Bind(wx.EVT_BUTTON, self.onButton2)
        
        if 0:
            ## no problem
            self.btn.Unbind(wx.EVT_BUTTON, handler=self.onButton2)
            self.btn.Unbind(wx.EVT_BUTTON, handler=self.onButton1)
        else:
            ## eversing the unbind order will break the linkage of the event chain
            self.btn.Unbind(wx.EVT_BUTTON, handler=self.onButton1)
            self.btn.Unbind(wx.EVT_BUTTON, handler=self.onButton2)
    
    def onButton1(self, evt):
        print(1)
        evt.Skip()
    
    def onButton2(self, evt):
        print(2)
        evt.Skip()

app = wx.App()
frm = wx.Frame(None)
frm.panel = TestPanel(frm)
frm.Show()
app.MainLoop()
```
</details>

swt2c:
When binding events to multiple methods and then unbinding them later,
in the same order they were bound, the wrong method would get unbound.

For example:
    self.btn.Bind(wx.EVT_BUTTON, self.onButton1)
    self.btn.Bind(wx.EVT_BUTTON, self.onButton1)
followed by:
    self.btn.Unbind(wx.EVT_BUTTON, handler=self.onButton2)
    self.btn.Unbind(wx.EVT_BUTTON, handler=self.onButton1)
works, but the reverse fails:
    self.btn.Unbind(wx.EVT_BUTTON, handler=self.onButton1)
    self.btn.Unbind(wx.EVT_BUTTON, handler=self.onButton2)

The reason is that the wxPython Disconnect() method called the wxWidgets
Disconnect() method with the userData parameter set to NULL.  In this
case, wxWidgets performs no filtering based on the userData parameter
and this could result in the wrong handler getting disconnected.

Fix this by setting the userData to a known value before calling
wxWidgets Disconnect() method so that it will disconnect the correct
handler.

This commit also adds a test that verifies the fix.

Fixes #2027.

### DIFF
--- a/etg/event.py
+++ b/etg/event.py
@@ -239,10 +239,13 @@ def run():
                         // is evaluated. (!!!)
                         if (PyObject_RichCompareBool(cb->m_func, func, Py_EQ) == 1) {
                             delete cb;
-                            entry->m_callbackUserData = NULL;
+                            // Set callback data to a known value instead of NULL to
+                            // ensure Disconnect() removes the correct handler.
+                            entry->m_callbackUserData = new wxObject();
                             // Now Disconnect should work
                             return self->Disconnect(id, lastId, eventType,
-                                                    (wxObjectEventFunction)&wxPyCallback::EventThunker);
+                                                    (wxObjectEventFunction)&wxPyCallback::EventThunker,
+                                                    entry->m_callbackUserData);
                         }
                     }
                     entry = self->GetNextDynamicEntry(cookie);
--- a/unittests/test_event.py
+++ b/unittests/test_event.py
@@ -241,6 +241,28 @@ class Events(wtc.WidgetTestCase):
 
 
 
+    def test_eventUnbinding5(self):
+        "Unbind in same order of binding"
+
+        class Frame(wx.Frame):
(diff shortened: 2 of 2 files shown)


RobinD42:
Merging. Ignoring timeouts in Linux CI.

PR #2036, opened 2021-11-27
buildtools : wxpysip to support sip version 6

bitranox:
buildtools : wxpysip to support sip version 6


swt2c:
Thanks, but I believe this is a duplicate of #1916.

bitranox:
@swt2c , yes, it seems like a duplicate, I am sorry. Please merge one of the two.
The next problem I ran into, is some compiler error, because the output of sip seems to be broken - I will write a report,
maybe You can point me in to the right direction. A quick skim over the code did not reveal where to change that ... 
bitranox

PR #2047, opened 2021-12-07
Use wx.DF_UNICODETEXT instead of wx.DF_TEXT

### LINKED ISSUE #2042: Looks like on Mac wx.TheClipboard.IsSupported doesn't return correct value
**Operating system**: macOS 11.4
**wxPython version & source**:       4.1.1
**Python version & source**:            3.7.0

Looks like "**wx.TheClipboard.IsSupported(wx.DataFormat(wx.DF_TEXT))**" always return false even if it does has text in clipboard. 

For example, first copy some text to clipboard; however the following code doesn't print anything 
```python
wx.TheClipboard.Open()
if wx.TheClipboard.IsSupported(wx.DataFormat(wx.DF_TEXT)):
    data = wx.TextDataObject()
    if wx.TheClipboard.GetData(data):
        print(data.GetText())
wx.TheClipboard.Close()
```
The following code print the text in clipboard correctly 
```python
wx.TheClipboard.Open()
data = wx.TextDataObject()
if wx.TheClipboard.GetData(data):
    print(data.GetText())
wx.TheClipboard.Close()
```
</details>

swt2c:
wx.DF_TEXT doesn't work on macOS and Unicode should be available under all platforms.

Fixes #2042

### DIFF
--- a/docs/sphinx/rest_substitutions/snippets/python/converted/wx.Clipboard.1.py
+++ b/docs/sphinx/rest_substitutions/snippets/python/converted/wx.Clipboard.1.py
@@ -6,7 +6,7 @@
 
 
         # Test if text content is available
-        not_empty = wx.TheClipboard.IsSupported(wx.DataFormat(wx.DF_TEXT))
+        not_empty = wx.TheClipboard.IsSupported(wx.DataFormat(wx.DF_UNICODETEXT))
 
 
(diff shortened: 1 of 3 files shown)


RobinD42:
Merging. Ignoring timeout in CI.

PR #2055, opened 2021-12-19
Don't let introspect.getCallTip warn

komoto48g:
This PR prevents DeprecationWarning about `inspect.formatargspec` deprecated since Python 3.5.

When the shell shows a calltip, the DeprecationWarning message is displayed every after the command is pushed.
We change the problematic part to the new method `inspect.signature`.


komoto48g:
I close this PR and make another one.

PR #2062, opened 2021-12-27
Fix additional Python 3.10 issues with AGW FlatNotebook

swt2c:


swt2c:
Don't merge yet - needs more fixes.

PR #2063, opened 2021-12-27
macOS: use default SDK to check if universal capable

swt2c:
@RobinD42 let me know what you think about this.  The reason the macOS CI was failing is because its default SDK (for some reason) is MacOSX10.15.sdk, but it has newer ones installed.  Thus, the build script was determining that universal was supported, but then tried to use it with the 10.15 SDK, which failed.  This solution seems like it should work?

### DIFF
--- a/.azure/ci-macos-job.yml
+++ b/.azure/ci-macos-job.yml
@@ -3,7 +3,7 @@ jobs:
   timeoutInMinutes: 90
   displayName: 'MacOS CI'
   pool:
-    vmImage: 'macOS-1015'
+    vmImage: 'macOS-11'
   strategy:
     matrix:
       Py37:
--- a/buildtools/build_wxwidgets.py
+++ b/buildtools/build_wxwidgets.py
@@ -63,6 +63,10 @@ def getXcodePaths():
     return [base, base+"/Platforms/MacOSX.platform/Developer"]
 
 
+def getDefaultSDKVersion():
+    return getoutput("xcrun --show-sdk-version")
+
+
 def getVisCVersion():
     text = getoutput("cl.exe")
     if 'Version 13' in text:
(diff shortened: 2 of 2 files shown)


RobinD42:
I'll experiment with it a little here and see how it goes. I think I tried something with `--show-sdk-path` at one point, but had some sort of problem with it.

RobinD42:
One potential problem: It's possible for `xcrun` to spit out a path with no version after the MacOS, like this:

```
 >  xcrun --show-sdk-path
/Applications/Xcode-13.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
```

There's an easy fix for that however, `xcrun` has a `--show-sdk-version` flag that will write the version number to stdout. In one case I also get some noise written to stderr so `getoutput` would also need to be tweaked to allow for getting just the stdout.

There's a bigger issue however. In cases like this where the default is the 10.15 SDK then there isn't a way to get a universal build even if there are capable SDKs available on the system. So we'll probably need a combination of the current approach with an additional optional flag for the wxpython build scripts that can be used to specify the SDK, and checking for the xcode default like in this PR if that flag isn't given.

I'll give it some more thought.

RobinD42:
Maybe a new `build.py` flag like: `--mac_sdk` with options "default", "best" or an explicit SDK path.

swt2c:
> Maybe a new `build.py` flag like: `--mac_sdk` with options "default", "best" or an explicit SDK path.
> 
> I'll work on that later today or tomorrow.

I guess my thinking on that was that the user could change the default SDK (I think by setting an environment variable?).  That way it would be set consistently for building wxWidgets + wxPython.

swt2c:
@RobinD42 see what you think about these latest changes.

RobinD42:
There's still a possible gap. If SDKROOT is set then `xcrun --show-sdk-version` may still print the default, or print an error, depending on the version of macOS or what SDK it's pointing to. So I think we'll still need a way to explicitly set the SDK on the build command-line, to handle the cases where trying to handle it automatically might fail. I've got that mostly implemented, I'll try to finish it up soon.

swt2c:
I ended up changing the macOS CI to use macOS 11, which resolved the immediate problem of CI not working on macOS.

PR #2071, opened 2022-01-04
Fix a bunch of test issues on GTK and OSX, plus a few non-test bugs

swt2c:


swt2c:
The tests all pass on wxGTK now, except for a couple of bugs that are fixed upstream in wxWidgets.  There's a couple that still fail on wxOSX (a segfault with wx.lib.agw.aui).

PR #2098, opened 2022-02-02
add Python 3.6 back

enkore:
It's kinda icky but 3.6 is still supported on some distros and 4.1.2 would have a lot of fixes from wx 3.1.5, so I think it'd be better if 4.1.2 is the last release for Py 3.6 instead of 4.1.1 from 2020.


swt2c:
My vote would be no.  Not that we won't fix bugs with 3.6, but we need to conserve our limited CI build time to supported Python releases.  Of course it's ultimately up to Robin.

PR #2111, opened 2022-02-24
MNT: Coerce argument types in call to DC.DrawText

pauldmccarthy:
A small and hopefully obvious fix to`wx.lib.stattext` which addresses the following error:

```
File "<envdir>/lib/python3.10/site-packages/wx/lib/stattext.py", line 306, in OnPaint                                                                                   
    dc.DrawText(line, x, y)                                                                                                                                                                    
TypeError: DC.DrawText(): arguments did not match any overloaded call:                                                                                                                         
  overload 2: argument 2 has unexpected type 'float'                                                                                                                                           
```

Without this patch, I'm experiencing this error on macOS 10.15 / wxpython 4.1.1.

Thanks!

### DIFF
--- a/wx/lib/stattext.py
+++ b/wx/lib/stattext.py
@@ -303,7 +303,7 @@ class GenStaticText(wx.Control):
                 x = width - w
             if style & wx.ALIGN_CENTER:
                 x = (width - w)//2
-            dc.DrawText(line, x, y)
+            dc.DrawText(line, int(x), int(y))
             y += h
 
 
@@ -319,4 +319,3 @@ class GenStaticText(wx.Control):
 
 
 #----------------------------------------------------------------------
-


wx/lib/stattext.py
```diff
@@ -303,7 +303,7 @@ def OnPaint(self, event):
                 x = width - w
             if style & wx.ALIGN_CENTER:
                 x = (width - w)//2
-            dc.DrawText(line, x, y)
+            dc.DrawText(line, int(round(x)), int(round(y)))
```
swt2c, line 306: To preserve existing behavior, I think it should just be: `dc.DrawText(line, int(x), int(y))`

pauldmccarthy: No problem - PR updated!

PR #2152, opened 2022-05-06
Fix CheckListBox GetSelections()

### LINKED ISSUE #2151: GetSelections() of CheckListBox is broken
**wxPython version & source**: self-built from master
**Python version & source**: stock Python 3.9

When calling `GetSelections()` on a (multiselect) `CheckListBox` object a `TypeError` is raised.

Steps to reproduce:

- Start with the CheckListBox example from the wxPython demo.
- Change creation of CheckListBox to use `style=wx.LB_EXTENDED` (for multi select)
- Add the following lines to the event handler `EvtListBox`: 
  ```
  indices = self.lb.GetSelections()
  self.log.WriteText(f"{indices=}")
  ```
- Run and select multiple elements in the box:

```
  File "[...]CheckListBox.py", line 37, in EvtListBox
    indices = self.lb.GetSelections()
TypeError: GetSelections(self, aSelections: object): not enough arguments
```

When passing an additional argument to `GetSelections()`, the number of selected items is returned instead of a list with the indices of the selected items.

`GetSelections()` in the base class `ListBox` still works as expected.

radoering:
Fixes #2151

I have no idea if this is the correct fix or if there is a better solution. At least it works for us.

### DIFF
--- a/etg/checklst.py
+++ b/etg/checklst.py
@@ -41,6 +41,15 @@ def run():
     c.find('Create').findOverload('wxString choices').ignore()
     c.find('Create').findOverload('wxArrayString').find('choices').default = 'wxArrayString()'
 
+    c.addCppMethod('wxArrayInt*', 'GetSelections', '()',
+        isConst=True, factory=True,
+        doc="Returns a list of the indices of the currently selected items.",
+        body="""\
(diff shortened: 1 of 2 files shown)


swt2c:
Thanks!

PR #2155, opened 2022-05-10
CI: save wheels as artifacts after building

swt2c:


swt2c:
I think the reason this fails is because saving artifacts doesn't support wildcards for filenames, so I would have to figure out the exact name of the wheel.

The main reason I was pursuing this was because the buildbot was broken, so I was going to save the wheels here so people could at least download them from somewhere.  Since you're back and fixed the buildbot, most of the motivation is gone.  Unless you'd prefer this approach and phase out the buildbot?

RobinD42:
At first I was thinking the extra artifacts would be good to have, but after some more thought I think it could lead to confusion. Not only for there being 2 sources, but also because there would be intermediate builds from branches available, and not just those made from the master branch.

In the long run I would like to phase out the buildbot, but I don't think we're there yet. With the buildbot there is still a bit more flexibility and control over what is built, what is installed in the build environments, etc.

PR #2161, opened 2022-05-16
BF: Sizer proportion parameter must be an integer

pauldmccarthy:
Howdy, this PR fixes a small bug in the `wx/lib/agw/aui/framemanager.py` - a floating point value is being passed as the `proportion` parameter to the `BoxSizer.Add()` method, which causes an error when trying to re-size a panel (on macOS):

```
  File "<env>/lib/python3.10/site-packages/wx/lib/agw/aui/framemanager.py", line 9285, in OnMotion
    self.OnMotion_Resize(event)                                                                
  File "<env>/lib/python3.10/site-packages/wx/lib/agw/aui/framemanager.py", line 9512, in OnMotion_Resize
    self.DoEndResizeAction(event)                                                              
  File "<env>/lib/python3.10/site-packages/wx/lib/agw/aui/framemanager.py", line 9134, in DoEndResizeAction
    return self.RestrictResize(clientPt, screenPt, createDC=False)                             
  File "<env>/lib/python3.10/site-packages/wx/lib/agw/aui/framemanager.py", line 9236, in RestrictResize
    self.Update()                                                                              
  File "<env>/lib/python3.10/site-packages/wx/lib/agw/aui/framemanager.py", line 6388, in Update
    self.DoUpdate()                                                                            
  File "<env>/lib/python3.10/site-packages/wx/lib/agw/aui/framemanager.py", line 6462, in DoUpdate
    sizer = self.LayoutAll(self._panes, self._docks, self._uiparts, False)                     
  File "<env>/lib/python3.10/site-packages/wx/lib/agw/aui/framemanager.py", line 6324, in LayoutAll
    uiparts = self.LayoutAddDock(cont, row, uiparts, spacer_only)                              
  File "<env>/lib/python3.10/site-packages/wx/lib/agw/aui/framemanager.py", line 5948, in LayoutAddDock
    uiparts = self.LayoutAddPane(dock_sizer, dock, pane, uiparts, spacer_only)                 
  File "<env>/lib/python3.10/site-packages/wx/lib/agw/aui/framemanager.py", line 5817, in LayoutAddPane
    sizer_item = cont.Add(horz_pane_sizer, pane_proportion,                                    
TypeError: Sizer.Add(): arguments did not match any overloaded call:                           
  overload 1: argument 1 has unexpected type 'BoxSizer'                                        
  overload 2: argument 1 has unexpected type 'BoxSizer'                                        
  overload 3: argument 2 has unexpected type 'float'                                           
  overload 4: argument 2 has unexpected type 'float'                                           
  overload 5: argument 1 has unexpected type 'BoxSizer'                                        
  overload 6: argument 1 has unexpected type 'BoxSizer'                                        
  overload 7: argument 1 has unexpected type 'BoxSizer'                                        
  overload 8: argument 1 has unexpected type 'BoxSizer'                                        
  overload 9: argument 1 has unexpected type 'BoxSizer'      
```

### DIFF
--- a/wx/lib/agw/aui/framemanager.py
+++ b/wx/lib/agw/aui/framemanager.py
@@ -5580,7 +5580,7 @@ class AuiManager(wx.EvtHandler):
         # finally, add the pane sizer to the dock sizer
         if pane.HasBorder():
             # allowing space for the pane's border
-            sizer_item = cont.Add(horz_pane_sizer, pane_proportion,
+            sizer_item = cont.Add(horz_pane_sizer, int(round(pane_proportion)),
                                   wx.EXPAND | wx.ALL, pane_border_size)
             part = AuiDockUIPart()
             part.type = AuiDockUIPart.typePaneBorder
@@ -5592,7 +5592,7 @@ class AuiManager(wx.EvtHandler):
             part.sizer_item = sizer_item
             uiparts.append(part)
         else:
-            sizer_item = cont.Add(horz_pane_sizer, pane_proportion, wx.EXPAND)
+            sizer_item = cont.Add(horz_pane_sizer, int(round(pane_proportion)), wx.EXPAND)
 
         return uiparts
 


RobinD42:
For Python < 3.10 it would truncate the float args without rounding. IOW, the equivalent of `int(f)` not `int(round(f))`. Is there a reason that rounding would be better in this case?

RobinD42:
For Python < 3.10 it would truncate the float args without rounding. IOW, the equivalent of `int(f)` not `int(round(f))`. Is there a reason that rounding would be better in this case?

pauldmccarthy:
@RobinD42 using `int(round(X))` is just my gut instinct when faced with this problem :) I'm not intimately familiar with the code, but it looks like this proportion value will range from 0 to 100000 (see [here](https://github.com/wxWidgets/Phoenix/blob/03499722b00261c1803a2c31978236b4c1b440a6/wx/lib/agw/aui/framemanager.py#L4564-L4566) and [here](https://github.com/wxWidgets/Phoenix/blob/03499722b00261c1803a2c31978236b4c1b440a6/wx/lib/agw/aui/framemanager.py#L8907-L8920)). 

So there's probably not much in it between `int(X)` versus `int(round(X))`? I'm happy to go with whatever you prefer.

RobinD42:
You're right. It probably doesn't make enough difference in this case to matter, and I think there have also been a couple other fixes that used `round` so I'm not going to be a stickler about it.

PR #2163, opened 2022-05-16
Autoconvert to wx.BitmapBundle

RobinD42:
Add a `convertFromPyObject` for wx.BitmapBundle so we can still pass a wx.Bitmap (or icon or image) where a wx.BitmapBundle is expected.

A few other wx.BitmapBundle related tweaks are also included.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,13 +18,13 @@ Pip:    ``pip install wxPython==4.1.2``
 
 New and improved in this release:
 
-* Yes, it's been a long time since the last release. I'm not dead, just on an
+* Yes, it's been a VERY long time since the last release. I'm not dead, just on an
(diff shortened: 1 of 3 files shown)


brendan-simon-indt:
Awesome Robin :)
I'm looking forward to trying it as soon as it is merged into `master` (as wxGlade doesn't work with the current latest snapshot `wxPython-4.1.2a1.dev5330` and have to use `wxPython-4.1.2a1.dev5310` in the meantime).

BrendanSimon:
Build of distributions failed.  Seems like the `bdist_wheel` completes ok, but the upload to the snapshots are fails (no valid filename - presumably doesn't exist) :(

brendan-simon-indt:
`wxGlade` `v1.0.4` is now working for me with `wxPython` `4.1.2a1.dev5355+6a5e93d0`

PR #2164, opened 2022-05-16
Update msvc detection

RobinD42:
Reduce dependency on distutils and fetch info about the installed MSVC compilers from setuptools. This will also switch the build to use Visual Studio 2017+ instead of trying to limit it to VS 2015.


RobinD42:
This is an alternative to #2805

PR #2177, opened 2022-06-05
Remove 3.6, add 3.10

oscargus:
Minor thing, but should go there at some stage.


oscargus:
(A bit of background: I had some issues with getting wxPython to run in the matplotlib tests, so I thought I'd check this out. At the end of the day, this was what was remaining.

Feel free to not merge this and leave the change to a regular developer.)

Issue #1582, opened 2020-03-26
Targetting MSYS2


sancelot:
**Operating system**:win10 / MSYS2
**wxPython version & source**:         4.0.7.post2
**Python version & source**:            distro python 3.8.0

**Description of the problem**:
Hi, 
I would need wxPython 4.0 ins MSYS2 
I tried compiling .
At first I would like to understand why it want to try to download and build waf since it is already in my system (and thus adapted to my system) ???

RobinD42:
> At first I would like to understand why it want to try to download and build waf

The `build.py` script expects by default to use specific versions of waf, sip, and doxygen. There is a `bin/waf-*` included in the source tarfile, but if it doesn't match the MD5 (for example, if line-endings were changed when it was unpacked) then it will download a fresh copy.

> Testing pyext configuration : Could not build python extensions

If you haven't already please take a look at the log from waf's configure step, probably `build/waf/3.8/config.log`. You should see there the C code it used to test if it's able to build Python extensions, and the errors that it ran into. That info may point you to the source of the problem.

stambaughw:
> The build.py script expects by default to use specific versions of waf, sip, and doxygen. There is a bin/waf-* included in the source tarfile, but if it doesn't match the MD5 (for example, if line-endings were changed when it was unpacked) then it will download a fresh copy.

@RobinD42 By specific versions do you mean some minimum version or an exact version?  The [KiCad project](https://kicad-pcb.org/) currently uses msys2 as our build system on windows so we've been shipping Python2 and wxWidgets on windows because Phoenix does not build on msys2 so I would really like to see this work completed.

@sancelot Do you have a public branch of your `build.py` changes on GitHub?  I would like to take a look at them if possible.

RobinD42:
Specific versions. See `build.py` starting around line 88. You can supply your own versions of those tools, if needed, by setting their paths in environment variables `SIP`, `WAF` and `DOXYGEN`.

RobinD42:
Also, in case it wasn't clear, if you build from a source tarball instead of a git checkout, then all you need is `waf`.

stambaughw:
[...]

PR #2200, opened 2022-07-02
Allows the environmental setting of CC_NAME to change compiler to gcc or clang

stahta01:
from the Windows default of msvc.

I am updating the MSys2 wxPython package and after the MSys2 waf package change I needed to create a patch that works with the new waf command. We use gcc and clang under windows to build the package. This commit allows the export of CC_NAME to pick gcc or clang as the compiler.

See [MSys2 MinGW PR](https://github.com/msys2/MINGW-packages/pull/11982)

Tim S.

### DIFF
--- a/buildtools/config.py
+++ b/buildtools/config.py
@@ -71,7 +71,10 @@ class Configuration(object):
     WXDLLVER = None
     # Version part of wxWidgets LIB/DLL names
 
-    COMPILER = 'msvc'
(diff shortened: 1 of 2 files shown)


stahta01:
This might close #1582

mmuetzel:
IIUC, you'd like to differentiate between a MSVC and a MinGW tool chain. Afaict, clang can be used as a MSVC compiler. Maybe consider renaming those variables to something more closely related to MinGW than the name of the compiler suite.

PR #2211, opened 2022-07-11
Simplify building with MinGW toolchain

mmuetzel:
IIUC, Phoenix is built with a MSVC toolchain on Windows by default.
While it is also possible to build it with a MinGW toolchain, this requires that some files of the build system are modified manually. The files that need changes include `wscript` and `buildtools/config.py`.

To simplify building with a MinGW toolchain a little bit, add a new option `--no_msvc` for `waf` that switches from a MSVC toolchain to a MinGW toolchain on Windows.
Without that option, the MSVC toolchain continues to be used on Windows. So by default, MSVC is used like before this change.

### DIFF
--- a/buildtools/config.py
+++ b/buildtools/config.py
@@ -142,12 +142,15 @@ class Configuration(object):
         self.finishSetup()
 
 
-    def finishSetup(self, wx_config=None, debug=None):
+    def finishSetup(self, wx_config=None, debug=None, compiler=None):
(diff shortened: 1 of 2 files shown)


RobinD42:
I'll merge this, but it would also be good to add a flag for build.py so it will also use MinGW to build wxWidgets, and also pass your new --no_msvc flag to waf.

PR #2215, opened 2022-07-25
IntCtrl: Change default colour to wx.NullColour

### LINKED ISSUE #2199: ``IntCtrl`` text colour doesn't respect system theme or enabled status
**Operating system**: Arch Linux
**wxPython version & source**: pypi v4.1.1
**Python version & source**: anaconda (v3.9)

I am developing a wxpython on a system with a dark theme (KDE as it happens) and I noticed that the default black for foreground doesn't show up particularly well against the dark grey background it has for me. I feel it would be more sensible to use whatever the default text foreground colour is for a ``TextCtrl`` rather than always selecting black. (Always defaulting to red for ``oob_color``, however, seems fine though.)

Additionally, the text colour doesn't change depending on whether or not the ``IntCtrl`` is enabled, unlike ``TextCtrl``. It'd be nice to have this feature. (Maybe a separate issue should be raised for this?)

alexdewar:
Fixes #2199.

The problem is that the default colour is currently wx.BLACK which will basically be invisible if the user has a dark theme. Use ``wx.NullColour`` instead so that it can just use the default text colour for the user's theme.

Note that this does not fix the second part of issue #2199 (the colours still don't change if the control is disabled) but I feel this should be a separate issue now.

### DIFF
--- a/wx/lib/intctrl.py
+++ b/wx/lib/intctrl.py
@@ -398,7 +398,7 @@ class IntCtrl(wx.TextCtrl):
                 name = "integer",
                 min=None, max=None,
                 limited = 0, allow_none = 0, allow_long = 0,
-                default_color = wx.BLACK, oob_color = wx.RED,
+                default_color = wx.NullColour, oob_color = wx.RED,
         ):
         """
         Default constructor


alexdewar:
Ping ping? Any chance of a review for this? :smile: 

It's a fairly trivial change but I think it's a good one!

Looking at the CI failures, it seems to have nothing to do with the PR and seems to be some generic Windows problem. It seems I don't have the authority to trigger a rerun of tests, but this might fix the issue....

PR #2216, opened 2022-07-26
Update wxWidgets to v3.2.0 (official release)

swt2c:


swt2c:
I thought the Windows build issue may have had something to do with setuptools versions, but it actually seems to be a regression in the Azure Pipelines image.  I reported it and they confirmed it's a bug.

PR #2220, opened 2022-08-06
Enable configuration of whether popup dialog should align center, right or left

jmoraleda:
This PR enables configuring horizontal alignment of the popup dialog with respect to the parent control in the python-only popup control.

The default now is center, which was the previous behavior, so this PR does not impact existing code.

However there are cases when one may prefer right or left alignment.  In particular right alignment makes sense when the control is wide and the pop up is narrow, so the pop up appears below the expand button.

### DIFF
--- a/wx/lib/popupctl.py
+++ b/wx/lib/popupctl.py
@@ -118,13 +118,16 @@ class PopupDialog(wx.Dialog):
         self.win.SetClientSize(self.content.GetSize())
         self.SetSize(self.win.GetSize())
 
-    def Display(self):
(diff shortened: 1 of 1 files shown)


jmoraleda:
Sorry for the dirty history. The second force-push was a mistake and the third one undoes it. Thus the last two "comparisons" are bogus showing lots of changed files are bogus. The first one, or the single commit, are the only one meaningful comparisons. This PR is very straight forward, modifying only a few lines in a single file.

jmoraleda:
Could someone please review this PR? It is only a handful of lines in a single file and I believe it is a useful feature. Thank you.

PR #2232, opened 2022-08-13
Make pip usage in wxget optional

StefanBruens:
As the code states, using pip to download is abusing it, and it is
just a fallback in case neither wget nor urllib works.


jmoraleda:
A comment about syntax: Is there an `is` missing in line 203? Should it say `if not success and pip is not None:`?

Metallicow:
@jmoraleda correctly corrected.
... Looks fine to me ...

PR #2252, opened 2022-08-29
OSX: remove check of SetFrontProcess() before App is fully initialized

### LINKED ISSUE #2250: MacOS isDisplayAvailable() and  Framework vs non-Framework build
**Operating system**:
**wxPython version & source**:     Github Master branch, MacOS     
**Python version & source**:         Anaconda especially, but potentially others?   

wxPython apps *claim* to need a Framework build.  Notably, with Anaconda Python, running `python wx_using_app.py` will give

    This program needs access to the screen. Please run with a
    Framework build of python, and only when you are logged in
    on the main display of your Mac

which comes from `core.py` and `isDisplayAvailable()` returning `False`.  

As described at https://stackoverflow.com/questions/72924604/how-to-run-wxpython-with-the-python-org-installer-on-macos-12-4-monterey/73386418#73386418, simply short-circuiting `isDisplayAvailable()` to be True does in fact work to display and run wx Apps, even without a framework build of Python.   In fact, doing

     import wx
     wx.PyApp.IsDisplayAvailable = lambda _: True

always allows the application to run and seems to run without problems.  

It appears that `isDisplayAvailable` (in `app_ex.cpp`) uses the deprecated functions `GetCurrentProcess()` and `SetFrontProcess()`, the second of which is returning an invalid value (-606) when being called as a non-Framework-built Python.

Could it be that this check of "can this app draw to the screen?" is being called too early, at least when not being called from a non-Framework build Python?   That is, this code in `isDisplayAvailable()` is being called before `applicationDidFinishLaunching()` is executed (from wxWidgets `src/osx/cocoa/utils.mm`) which does do the

	    [NSApp activateIgnoringOtherApps: YES];

that appears to be needed to permit drawing to the screen.

I'm not sure I know what the best solution here is, but is it actually still needed to have `isDisplayAvailable()` actually test with
the deprecated `SetFrontProcess()`?     I get that the screen might not actually be available, though that would imply someone had ssh-ed in or something, which seems pretty unusual on a Mac. 

Presumably, if the screen could not be drawn, one of those wxWidgets calls to "[NSApp ....]" would fail.  Perhaps that could be captured later in the startup process and fail only if the application did not finish launching?

newville:
Fixes #2250 

This simply removes the calls to `GetCurrentProcess()` and `SetFrontProcess()` from `IsDisplayInitialized()`.   These two functions are deprecated, and `SetFrontProcess()` returns a value of -606 for non-framework builds of Python (notably including Anaconda Python), which then falsely raises an exception that the Application cannot draw to the screen.

With current wxWidgets on MacOSX, 

    [NSApp activateIgnoringOtherApps: YES];

and related calls to `[NSApp ...]` are called by `applicationDidFinishLaunching()`, which allows drawing to the screen.  This method is not called from wxPython until `CallOnInit()` time, which is much later in the initialization step than `IsDisplayInitialized()`.   I do not know if this approach is acceptable for cases where the application truly does not have write access to the screen, but relying on calls to deprecated functions is probably not that reliable anyway, and it appears that  `SetFrontProcess()` is often giving "false negative" results.

This fix would also satisfy the long-standing (getting close to 8 years old!) issue raised with Anaconda Python:  https://github.com/ContinuumIO/anaconda-issues/issues/199

### DIFF
--- a/src/app_ex.cpp
+++ b/src/app_ex.cpp
@@ -335,11 +335,7 @@ bool wxPyApp::IsDisplayAvailable()
         //
         // [NSRunningApplication activateIgnoringOtherApps: YES]  ??
         // 
-        if (GetCurrentProcess(&psn) < 0 || SetFrontProcess(&psn) < 0) {
-            rv = false;
-        } else {
-            rv = true;
-        }
+        rv = true;
     }
     return rv;
 #endif


RobinD42:

https://discuss.wxpython.org/t/wxpython-stops-working-on-macos-12-6-1-framework-build-error/36261/2

PR #2285, opened 2022-09-18
Patch out GTK specific function from wx 3.2.1

marekr:
This is a new GTK specific function added in wx 3.2.1 that results in build errors on Windows due to it's inclusion in the interface header but not windows itself.

### DIFF
--- a/etg/app.py
+++ b/etg/app.py
@@ -212,6 +212,12 @@ def run():
         #endif
         """)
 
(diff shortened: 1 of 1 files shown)


swt2c:
This can't be merged as-is because it breaks building with wx 3.2.0.

swt2c:
Applied in https://github.com/wxWidgets/Phoenix/commit/3812a286c4368265bf77aa03161c51f770202ce6.

PR #2298, opened 2022-10-25
Add CI for Python 3.11

swt2c:


swt2c:
Looks like Python 3.11 is not quite yet added to the Azure images.

Issue #2043, opened 2021-12-04
Crash when accessing wx.stc.StyledTextCtrl.DropTarget.Data


komoto48g:
**Operating system**: Windows 10
**wxPython version & source**: 4.1.1 (pypi)
**Python version & source**: 3.8.6

**Description of the problem**: This bug is the direct cause of the previously opened issue #1959 
Accessing the `Data` of `wx.stc.StyledTextCtrl.DropTarget` property will crash, but other properties are no problem.
I will raise this as a new issue and close #1959.

<details>
  <summary>Code Example (click to expand)</summary>

```python
import wx
import wx.stc

app = wx.App()
frm = wx.Frame(None)
text = wx.stc.StyledTextCtrl(frm)
print("text.DropTarget =", text.DropTarget)
print("text.DropTarget.Data =", text.DropTarget.Data) # crash here
print("done")
```
</details>

swt2c:
Hmm.  Possibly Windows-specific?  I don't see on Linux.  Does this happen on 4.0.7 for you also?

komoto48g:
Thank you for checking this issue. This occurs also for wx 4.0.7 on Windows 10.
Seems like a Windows-specific DnD (shell extension) problem...

swt2c:
Can you get a stacktrace for the crash?

komoto48g:
No, unfortunately.
The error occurs inside the built-in method `DropTarget.GetData or Data`.

RobinD42:
This issue has been mentioned on **Discuss wxPython**. There might be relevant details there:

https://discuss.wxpython.org/t/disabling-drag-and-drop-in-a-richtextctrl/35795/2

Issue #1959, opened 2021-05-06
FillingFrame HARD-CRASH when trying to expand the tree of EditWindow.


komoto48g:
**Operating system**: Windows 10 x64
**wxPython version & source**: 4.1.1 msw (pypi installed)         
**Python version & source**: Python 3.8.6            

**Description of the problem**: FillingFrame crashes when trying to expand the tree of the EditWindow object.

<details>
  <summary>Code Example (click to expand)</summary>

```python
import wx
from wx.py.filling import FillingFrame
from wx.py.editwindow import EditWindow

if __name__ == "__main__":
    app = wx.App()
    frm = wx.Frame(None)
    frm.panel = EditWindow(frm)
    frame = FillingFrame(frm, rootObject=frm.panel)
    frame.Show()
    frm.Show()
    app.MainLoop()
```
</details>

I'm not sure why, but as a little clue, add `frm.panel.SetDropTarget(None)` and it won't crash.

komoto48g:
The direct cause of this problem is described in #2043

PR #2314, opened 2022-12-04
Reduce FillingTree overheads

komoto48g:
This PR fixes FillingTree overheads and adds some customization features.

Prevent crash when referencing `DropTarget` (Windows bug fix) cf. #1959, #2043.
Suppress unnecessary deprecation warning messages.
Provide some customizable options:
- filter function of tree items
- formt function for value text

Reduce some overheads:
- Prevent call `display` => `addChildren` (=> `DeleteChildren`) => `objGetChildren`. Use`updateChildren`.
- Prevent call `objHasChildren` => `objGetChildren`. Use type check only.

The following video clip shows the effect of the fix.
Before:

https://user-images.githubusercontent.com/83063554/205479618-10afa088-697d-4569-af84-a3c74aa43cbe.mp4

After:

https://user-images.githubusercontent.com/83063554/205479629-903c7f7b-5dfe-4bc1-8bde-0f52a05adfa7.mp4

### DIFF
--- a/wx/py/filling.py
+++ b/wx/py/filling.py
@@ -106,10 +106,7 @@ class FillingTree(wx.TreeCtrl):
 
     def objHasChildren(self, obj):
         """Return true if object has children."""
-        if self.objGetChildren(obj):
-            return True
-        else:
-            return False
+        return type(obj) not in COMMONTYPES
 
     def objGetChildren(self, obj):
         """Return dictionary with attributes or contents of object."""
--- a/wx/py/filling.py
+++ b/wx/py/filling.py
@@ -125,6 +125,9 @@ class FillingTree(wx.TreeCtrl):
                 d[key] = obj[n]
         if otype not in COMMONTYPES:
             for key in introspect.getAttributeNames(obj):
+                if wx.Platform == '__WXMSW__':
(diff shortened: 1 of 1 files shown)


wx/py/filling.py
```diff
@@ -106,10 +106,7 @@ def OnItemActivated(self, event):
 
     def objHasChildren(self, obj):
         """Return true if object has children."""
-        if self.objGetChildren(obj):
-            return True
-        else:
-            return False
+        return type(obj) not in COMMONTYPES
```
komoto48g, line 109: Looking at `objGetChildren`, do I need to add the following?
```python
return ('BTrees' in six.text_type(otype) and hasattr(obj, 'keys') or ...
```
I don't know what `Btrees` type is.

komoto48g:
I think the new `updateChildren` that I added can be integrated into `addChildren` by the last fix.

L230-L255 is aggressive, I guess there is a better way to refresh the roots object.

komoto48g:
Reading
> "https://wxpython.org/pages/contributor-guide/"
> 5. Do not put more than one fix/feature in the same branch or PR. Make a new distinct branch and PR for each of them.
I think I should have created separate PRs for this.
I will close this and make a new PR.

PR #2360, opened 2023-03-04
Fix blurry text in AuiNotebook tab.

### LINKED ISSUE #2359: wx.lib.agw.aui.AuiNotebook tab text blurry
**Operating system**: MacOS (13.2.1)
**wxPython version & source**:        4.2.0
**Python version & source**:            3.9.5

The text on notebook tab is blurry (e.g, AUI.py in demo).

Screenshot 2023-03-01 at 10 54 53 AM

I think the problem is in the following line
https://github.com/wxWidgets/Phoenix/blob/986d79b94eed8984a7e273b34c5fdf57d266f60a/wx/lib/agw/aui/auibook.py#L1465

if we change it to (i.e., to apply the scaling factor from `raw_dc`), the problem goes away
```python
bmp = wx.Bitmap(self._rect.GetWidth(), self._rect.GetHeight(), raw_dc)
```
Screenshot 2023-03-01 at 10 54 08 AM

tianzhuqiao:
Fixes #2359

### DIFF
--- a/wx/lib/agw/aui/auibook.py
+++ b/wx/lib/agw/aui/auibook.py
@@ -1462,7 +1462,7 @@ class AuiTabContainer(object):
         button_count = len(self._buttons)
 
         # create off-screen bitmap
-        bmp = wx.Bitmap(self._rect.GetWidth(), self._rect.GetHeight())
+        bmp = wx.Bitmap(self._rect.GetWidth(), self._rect.GetHeight(), raw_dc)
         dc.SelectObject(bmp)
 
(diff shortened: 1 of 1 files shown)


jmoraleda (approved):
Looks good to me. In fact the Bitmap constructor that takes a display context as argument says in its help that this usage is exactly its purpose: "Create a bitmap compatible with the given DC, inheriting its magnification factor." https://docs.wxpython.org/wx.Bitmap.html#wx.Bitmap.__init__

Issue #2038, opened 2021-11-29
On Python 3.10: TypeError: 'float' object cannot be interpreted as an integer


cykerway:
**Operating system**: Fedora 35
**wxPython version & source**:         4.0.7 
**Python version & source**:           3.10 

**Description of the problem**:

After install wxpython using:

    dnf install python3-wxpython4

The example code throws:

>   TypeError: 'float' object cannot be interpreted as an integer

There is something very interesting: The fc35 package leaves its files at `/usr/lib64/python3.10/site-packages`. However, I didn't find 3.10 a supported Python version in wxpython `setup.py` (version 4.0.7, but same in master). Would you please confirm if wxpython will work with Python 3.10 on the Linux platform (Fedora, in specific)?

-   If answer is "yes", then what caused the exception?

-   If answer is "no", then it looks like Fedora maintainers are making a mistake by giving a 3.10 package? I can forward the issue there instead.

<details>
  <summary>Code Example (click to expand)</summary>

```python
#!/usr/bin/env python3

from wx.lib.agw.pygauge import PyGauge
import wx

class MyFrame(wx.Frame):
    def __init__(self, *args, **kw):
        super().__init__(*args, **kw)
        self.gauge = PyGauge(self, range=128)

app = wx.App()
MyFrame(None).Show(True)
app.MainLoop()

```
</details>

swt2c:
Thanks for the report.  Officially, I suppose wxPython doesn't support Python 3.10 *yet* but the Fedora package should be mostly working.  In this case, you've found a Python 3.10 bug with PyGauge.  I'll submit a fix for it (and let me know if you want it fixed in Fedora, too).

cykerway:
Thanks for quick response. I believe you can find more such bugs in `ListCtrl`, etc. It's somewhat ubiquitous because the float is introduced by a division `/`, which is everywhere in wxpython widgets source code. In Python 3.x a division always gives a float, but this assignment works with Python 3.9...

I have compared the shipped files in old and new RPMs. They turn out to be almost exactly the same (except for some build IDs). This implies the problem doesn't like in `.py` files but somewhere more fundamental. For this specific thread, I'm wondering why assigning a float to an int would work when using Python 3.9 but not 3.10.

swt2c:
[...]

PR #2387, opened 2023-05-01
agw: Fix TypeError caused by floats with Python 3.10 extensions.

Apteryks:
This fixes the following error:

   File "/lib/python3.10/site-packages/wx/lib/agw/pygauge.py", line 380, in OnPaint
       dc.DrawText(drawString, textXPos, textYPos)
   TypeError: DC.DrawText(): arguments did not match any overloaded call:
     overload 2: argument 2 has unexpected type 'float'
   TimeLeft: 3.0

Visible when using Python 3.10 or newer.

See: https://github.com/wxWidgets/Phoenix/issues/2038#issuecomment-1530004477.


jmoraleda:
Looks good to me.

PR #2395, opened 2023-05-22
fix TypeError in wx.lib.agw.flatnotebook for py3.10+ (included #2210)

### LINKED ISSUE #2210: wx.lib.agw.flatnotebook raises error on changing tab
**Operating system**: Ubuntu 22.04
**wxPython version & source**: 4.0.7 gtk3 (phoenix) wxWidgets 3.0.5 (installed with `apt install python3-wxgtk4.0`)
**Python version & source**: Python 3.10.4, stock

Changing tabs in FlatNotebook can raise an exception:
```
  File "/usr/lib/python3/dist-packages/wx/lib/agw/flatnotebook.py", line 5205, in OnPaint
    renderer.DrawTabs(self, dc)
  File "/usr/lib/python3/dist-packages/wx/lib/agw/flatnotebook.py", line 2360, in DrawTabs
    self.DrawTab(pc, dc, posx, i, tabWidth, tabHeight, pc._nTabXButtonStatus)
  File "/usr/lib/python3/dist-packages/wx/lib/agw/flatnotebook.py", line 2783, in DrawTab
    x_rect = wx.Rect(tabCloseButtonXCoord, tabCloseButtonYCoord, 16, 16)
TypeError: Rect(): arguments did not match any overloaded call:
  overload 1: too many arguments
  overload 2: argument 1 has unexpected type 'float'
  overload 3: argument 1 has unexpected type 'float'
  overload 4: argument 1 has unexpected type 'float'
  overload 5: argument 1 has unexpected type 'float'
  overload 6: argument 1 has unexpected type 'float'
```

The problem is that in https://github.com/wxWidgets/Phoenix/blob/master/wx/lib/agw/flatnotebook.py, a number of calculations use division, but do not explicitly convert the result to an integer. This is probably a holdover from Python2, where integer divided by an integer yielded an integer; in Python3 the result is always a float.

mprosperi:
fix for several float-int TypeError with python 3.10+ in wx.lib.agw.flatnotebook

Fixes #2210

### DIFF
--- a/wx/lib/agw/flatnotebook.py
+++ b/wx/lib/agw/flatnotebook.py
@@ -2232,7 +2232,7 @@ class FNBRenderer(object):
                 return
 
         # Get the text height
-        tabHeight = round(self.CalcTabHeight(pageContainer))
+        tabHeight = int(self.CalcTabHeight(pageContainer))
         agwStyle = pc.GetParent().GetAGWWindowStyleFlag()
 
         # Calculate the number of rows required for drawing the tabs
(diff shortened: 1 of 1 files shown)


swt2c:
Hi @mprosperi, thanks for the PR!  I think, however, that `int()` is preferred to `round()` in this case because the existing behavior on Python < 3.10 is to silently truncate floats.  So I think we should use `int()` to match existing behavior.

PR #2447, opened 2023-08-17
Fix AuiManager pane minimizing issue.

### LINKED ISSUE #2361: wx.lib.agw.aui Minimize pane issue
**Operating system**: MacOS 13.2.1
**wxPython version & source**:       4.2.0
**Python version & source**:            3.9.5

When drag a notebook page and dock it somewhere, it will not be able to be minimized. 

aui_minimize_pane_fail

The issues comes from the discrepancy between the paneinfo in `AuiManager._panes` and `AuiManager._uiparts`.
```python
def OnLeftUp_DragFloatingPane(self, eventOrPt):
        ...
        # here we change the paneinfo in _panes, but Update() may not be called

        # reorder for dropping to a new notebook
        # (caution: this code breaks the reference!)
        tempPaneInfo = self.CopyTarget(paneInfo)
        self._panes.remove(paneInfo)
        self._panes.append(tempPaneInfo)

        if ret:
            self.Update()

        if tempPaneInfo.IsFloating():
            self.SnapPane(tempPaneInfo, tempPaneInfo.floating_pos, tempPaneInfo.floating_size, False)
    ...
```
Looks like if we move the "reorder" part in the `if ret` section, the problem goes away.
```python
def OnLeftUp_DragFloatingPane(self, eventOrPt):
        ...
        if ret:
            # reorder for dropping to a new notebook
            # (caution: this code breaks the reference!)
            tempPaneInfo = self.CopyTarget(paneInfo)
            self._panes.remove(paneInfo)
            self._panes.append(tempPaneInfo)

            self.Update()

            if tempPaneInfo.IsFloating():
                self.SnapPane(tempPaneInfo, tempPaneInfo.floating_pos, tempPaneInfo.floating_size, False)
    ...
```
aui_minimize_pane_pass

tianzhuqiao:
Fixes #2361

### DIFF
--- a/wx/lib/agw/aui/framemanager.py
+++ b/wx/lib/agw/aui/framemanager.py
@@ -9465,17 +9465,17 @@ class AuiManager(wx.EvtHandler):
         elif self._has_maximized:
             self.RestoreMaximizedPane()
 
-        # reorder for dropping to a new notebook
-        # (caution: this code breaks the reference!)
-        tempPaneInfo = self.CopyTarget(paneInfo)
-        self._panes.remove(paneInfo)
-        self._panes.append(tempPaneInfo)
-
         if ret:
+            # reorder for dropping to a new notebook
+            # (caution: this code breaks the reference!)
+            tempPaneInfo = self.CopyTarget(paneInfo)
+            self._panes.remove(paneInfo)
+            self._panes.append(tempPaneInfo)
+
             self.Update()
 
-        if tempPaneInfo.IsFloating():
-            self.SnapPane(tempPaneInfo, tempPaneInfo.floating_pos, tempPaneInfo.floating_size, False)
+            if tempPaneInfo.IsFloating():
+                self.SnapPane(tempPaneInfo, tempPaneInfo.floating_pos, tempPaneInfo.floating_size, False)
 
         self.HideHint()
         ShowDockingGuides(self._guides, False)


jmoraleda:
@tianzhuqiao If you have a chance, could you please test if making the change in this image --removing three lines and making all platforms use `wx.CallAfter`-- also fixes your original issue, instead of reordering the lines?

image

I've been testing this change for a while before making a PR for it and I think it addresses multiple drawing and flickering issues and in particular I think the issue you report might also be addressed by it, but I don't have a Mac to test on. Thank you.

tianzhuqiao:
@jmoraleda, thanks, just tested it on macOS, it shows the same issue with the change (always use wx.CallAfter)

jmoraleda:
> @jmoraleda, thanks, just tested it on macOS, it shows the same issue with the change (always use wx.CallAfter)

Thank you for looking into it. Sorry I couldn't help.

Metallicow:
Kind of a touchy issue when you have to write something like... `if __WXGTK__ in wx.PlatformInfo` as I thought it might help years also and in some spots I wound back up at a conclusion as to WHY? would I do this if I had already commented out that line before on another linuxbox and just told it to wxCallAfter... Not sure but this may be what Robin was refering to as cruft when Phoenix was first RELz in the early versions. Might ask Robin... Is this logic really cruft?

In my newer WIPz "aui" rewrite, I have removed all offending platforms lines and started new. If it doesn't work, then I'll deal with specifics after its released if at all. ...Not monkeyjump all over the place into a million linux flavors. Maybe something such as the fabled focus event fix could screw something like this up also... idunno at this point. 

... would `If __GTK__("ubuntu 1947")` type of code might work better..? We have absolutely no reference to any specific os system it seems at this point...

Metallicow:
ThankYouKubuntu1984

PR #2453, opened 2023-08-29
Allow overriding wxSizer.InformFirstDirection()

### LINKED ISSUE #2452: InformFirstDirection() cannot be overridden in Python but a custom Sizer is not possible without it
**Operating system**: Ubuntu 20.04, Windows 10
**wxPython version & source**: 4.2.1
**Python version & source**: 3.8.10 (Ubuntu), 3.10.0 (Windows)

`InformFirstDirection()` must be overridden by classes that derive from Sizer in order to properly implement any non-trivial sizer, as indicated by this comment:

https://github.com/wxWidgets/wxWidgets/blob/be772c284f20969e7c2a5b62d3d43db56cdc5a6b/include/wx/sizer.h#L636-L642

However, wxPython forbids this method from being overridden in Python classes that derive from Sizer. The only methods that can be overridden are `CalcMin()` and `RepositionChildren()`:

https://github.com/wxWidgets/Phoenix/blob/e4fd9a3e047a4f2e29f2ca2839feb8e6134f706c/etgtools/tweaker_tools.py#L327-L333

This means it is not possible to implement a non-trivial custom Sizer in Python. Even the simple BoxSizer and WrapSizer need `InformFirstDirection()` to properly calculate their size.

I am trying to implement a custom Sizer in Python similar to WrapSizer that uses a box packing algorithm to more efficiently layout its children. Unfortunately, without this method being virtual, it is impossible to make it work properly.

(As a workaround I am having my custom sizer store a reference to its containing ScrolledPanel so it can call `GetClientSize()` on it during its `CalcMin()`. This is an ugly hack that should not be necessary.)

swt2c:
It seems this is required in order to implement a non-trivial custom Sizer.

Fixes #2452.

Fixes #NNNN

### DIFF
--- a/etgtools/tweaker_tools.py
+++ b/etgtools/tweaker_tools.py
@@ -434,6 +434,10 @@ def fixSizerClass(klass):
     removeVirtuals(klass)
     klass.find('CalcMin').isVirtual = True
     klass.find('RepositionChildren').isVirtual = True
+    try:
+        klass.find('InformFirstDirection').isVirtual = True
+    except extractors.ExtractorError:
+        pass
 
     # in the wxSizer class it is pure-virtual
     if klass.name == 'wxSizer':


Metallicow:
haha yep, I need to **_Force_** something...

... better remember that for latter days ...

PR #2480, opened 2023-11-15
Add support to python only pdfviewer for displaying pdf files where not all pages are the same size

jmoraleda:
The current PDF viewer assumes all pages in the PDF document being displayed are of the same size. When this is not the case the PDF does not display correctly.

This PR refactors the viewer so that PDF documents containing pages of different sizes display correctly.

I have thoroughly tested this PR in GTK and MSW and have found it to produce the correct results and have performance comparable to the previous version including for larger documents.

### DIFF
--- a/wx/lib/pdfviewer/viewer.py
+++ b/wx/lib/pdfviewer/viewer.py
@@ -23,11 +23,8 @@
 This module provides the :class:`~wx.lib.pdfviewer.viewer.pdfViewer` to view PDF
 files.
 """
-
-import sys
-import os
-import time
-import types
+import bisect
+import itertools
 import copy
 import shutil
 from six import BytesIO, string_types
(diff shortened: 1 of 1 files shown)


jmoraleda:
See review of this PR in https://github.com/wxWidgets/Phoenix/pull/2492#issuecomment-1858873301

jmoraleda:
I wonder if we could merge this in. 

It has been reviewed thoroughly at See discussion and review of this PR in https://github.com/wxWidgets/Phoenix/pull/2492#issuecomment-1858873301. That was an alternative PR for the same feature, and as you can see on the comments, all reviewers including the proponent of that PR agreed that this one was a better implementation.

Furthermore it has been thoroughly tested for several months by many people as part of the wxPython-Meticy release:  https://pypi.org/project/wxPythonMeticy/

jmoraleda:
@Infernio I noticed you worked on the python only pdfviewer recently. Could you by any chance test this PR and provide feedback? Thank you!

Infernio:
I don't use the pure Python one, judt the PyMuPDF one. That said, from looking over the commit, it seems fine.

PR #2488, opened 2023-11-27
fixed 'str' to 'wxString' converstation, when emoji is inside string

### LINKED ISSUE #2446: 'str' to 'wxString' converstation is broken, when emoji is inside string
**wxPython version & source**: wxPython 4.2.1 from pypi
**Python version & source**: Python 3.11

Whem some special emojis like 🔄 are inside a python string, that is converted to an wxString (internally in wxPython), the last character of the string is lost. It is also not shown in the gui.

The changes in https://github.com/wxWidgets/Phoenix/commit/29520a122a5a7c34cbd9951102562297a6b5afb5 creates the problem. I compiled it with and without the changes and only with the changes the error occures. I guess that `PyUnicode_GET_LENGTH` is not an 1:1 replacement for `PyUnicode_GET_SIZE`.

I created the following small example to illustrate the problem.

```python
import wx

app = wx.App(False)
txt = wx.Button()

def test_string(input):
    txt.SetLabel(input)
    output = txt.GetLabel()
    print(f"{input=}, {output=}")

test_string("asdf")  # Output is: "asdf" -> ok
test_string("⟳ asdf")  # Output is: "⟳ asdf" -> ok
test_string("🔄 asdf")  # Output is: "🔄 asd" -> error: last char is missing
test_string("🔄🔄 asdf")  # Output is: "🔄🔄 as" -> error: last two chars are missing
test_string("🔄")  # Output is: "\ud83d" -> error
```

timrid:
Fixes #2446

### DIFF
--- a/src/arrays.sip
+++ b/src/arrays.sip
@@ -59,7 +59,7 @@
             }
             PyErr_Clear();
             wxString string;
-            size_t len = PyUnicode_GET_LENGTH(item);
+            size_t len = PyUnicode_AsWideChar(item, NULL, 0);
             if (len) {
                 wxPyUnicode_AsWideChar(item, wxStringBuffer(string, len), len);
             }
--- a/src/string.sip
+++ b/src/string.sip
@@ -45,7 +45,7 @@
             }
         }
         *sipCppPtr = new wxString();
-        size_t len = PyUnicode_GET_LENGTH(uni);
+        size_t len = PyUnicode_AsWideChar(uni, NULL, 0);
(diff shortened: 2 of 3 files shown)


swt2c:
Thanks for investigating and for the fix.  I have a very slight worry since this use of `PyUnicode_AsWideChar` is only documented in the comments (it doesn't seem to be in the HTML documentation), but it seems that existing cpython code uses it, so it's probably OK.

PR #2492, opened 2023-12-02
Fix viewer.py Adapt page size on the fly if necessary

alexawake:
It fixes the issue of documents with page sizes different from the first one, these might not be viewed correctly afterwards since the viewer assumes all pages being the size of the first page. This is a typical scenario in documents where a first page with extra information is added, for instance on exporting this PDF https://eudml.org/doc/202780
(Access to full text + Export)

Related with pull request #2480 "Add support to python only pdfviewer for displaying pdf files where not all pages are the same size"

but with the adaption solution it is not needed that all pages of the document are computed in advance

### DIFF
--- a/wx/lib/pdfviewer/viewer.py
+++ b/wx/lib/pdfviewer/viewer.py
@@ -316,6 +316,14 @@ class pdfViewer(wx.ScrolledWindow):
 
     # This section is concerned with rendering a sub-set of drawing commands on demand
 
+    def CheckPageDimensions(self, width, height):
+        # check that document page width and height fits the page
+        # for example for documents with pages of different sizes, adjust it to the maximum
+        if width > self.pagewidth or height > self.pageheight:
+            self.pagewidth  = max(width, self.pagewidth)
+            self.pageheight = max(height, self.pageheight)
+            self.CalculateDimensions ()
+
     def CalculateDimensions(self):
         """
         Compute the required buffer sizes to hold the viewed rectangle and
@@ -506,7 +514,6 @@ class mupdfProcessor(object):
             page = self.pdfdoc.loadPage(0)
         self.pagewidth = page.bound().width
         self.pageheight = page.bound().height
-        self.page_rect = page.bound()
         self.zoom_error = False     #set if memory errors during render
 
     def DrawFile(self, frompage, topage):
@@ -522,6 +529,10 @@ class mupdfProcessor(object):
             page = self.pdfdoc.load_page(pageno)
         except AttributeError: # old PyMuPDF version
             page = self.pdfdoc.loadPage(pageno)
+
+        # adapt on the fly document page dimensions if necessary
+        self.parent.CheckPageDimensions (page.bound().width, page.bound().height)
+
         matrix = fitz.Matrix(scale, scale)
         try:
             try:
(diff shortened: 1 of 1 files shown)


alexawake:
Hi Jorge, testing your approach #2480 "choosing a bounding box for each page that exactly fits the page" with the example pdf, I see that the zoom changes on each page in order to do that fit of the page. But I don't think that this decision should be taken automatically.  If the PDF is designed with pages of different sizes for whatever reason this should be kept in that way.  Both Chrome and Edge browsers keep the different sizes of the pages in the example pdf,  the only difference with my approach is that using #2492 the pages are not centered, this could be fixed.

jmoraleda:
Hi Alejandro,

Thank you for taking the time to look at this.

> I see that the zoom changes on each page in order to do that fit of the page.

In PR #2480 The zoom changes automatically when the pdfviewer is set to "fit page width"  image (the initial setting of the pdfviewer widget) or "fit page height" image . In that case, the viewer adjusts the zoom for each page so the page fits the width or height as intended. When the pdfviewer is set to display at a constant zoom image then the zoom stays at the target zoom and each page is scaled to that level of zoom.

IMHO this is the intended and most correct behavior.

In all cases the page gap is adjusted to not leave extraneous space between pages. 

If other people are interested in this topic, I think it would be helpful to get them to test both approaches and get their opinion. Also, if anyone has real-world examples of pdf's with different page sizes that can be made publicly available, that would be helpful. (I have such examples, hence the motivation for my PR, but they contain private patient data and I cannot share them)

alexawake:
Hi Jorge, 
I've taken a deeper look into your change and the whole viewer.py in general can tell you following.
- your approach getting in advance all page sizes is the correct one:
 I've seen that takes almost no time even in huge documents, for instance in one about 1400 pages. But precisely testing that document it happens that if one page has issues (pdf corrupt) then an exception prevents to show anything of the document. It can be easily fixed by adding one more except in the lines
```
        """ Return width, height for the page """
        try:
            page = self.pdfdoc.load_page(pageNum)
        except AttributeError: # old PyMuPDF version
            page = self.pdfdoc.loadPage(pageNum)
        except: # otherwise if PDF is corrupt at some page
            return (100, 100) # for example or any other value

```
In general, not only your change but in all try / except AttributeError in viewer.py a "except:" line should be added with the right action, for instance in RenderPage it could be simply return

```
    def RenderPage(self, gc, pageno, scale=1.0):
        " Render the set of pagedrawings into gc for specified page "
        try:
            page = self.pdfdoc.load_page(pageno)
        except AttributeError: # old PyMuPDF version
            page = self.pdfdoc.loadPage(pageno)
        except: # otherwise if for example PDF is corrupt at some page
            return

```
regarding the adaption to the zoom on each page, testing your example pdf in other viewers all them in "page width mode" take the maximum width in the document and show all pages centered to it, which I think is the right approach.

- testing PyPDF2 capability
viewer.py dedicate more that 2 thirds of the code to supporting this library but : it is outdated, not working anymore due to deprecated calls or members
After fixing the deprecated stuff, I've seen that the viewer actually cannot render almost anything from a simple PDF.
I can fully understand the whish to work and improve the viewer with this full open library and withtout C modules but 
in any case, even if it were comparable with MuPDF, I would separate these two different approaches in two files, for example
          pdfViewerMuPDF.py
          pdfViewerPyPDF2.py
and if desired another module could do the "module switch" between both 

personally by now I will only work on  pdfViewerMuPDF.py and will publish it in my fork when finished.

jmoraleda:
Hello Alejandro,

Thank you again for your time. I am glad we have reached consensus. I wonder what is the best link to move/copy the discussion to PR #2480 since we have agreed that is the approach we would like to merge.

#### Being able to handle partially broken documents
I think this would be a good feature. And from your comments, it seems its implementation is straight forward. I can add support for it in PR #2480  before it gets merged or later in a separate PR if it gets merged before then. Could you add there a link to the broken document(s) you used for testing this, if they are public, so I can test this new feature when the user navigates to broken pages, or scrolls around broken pages? Thank you.

#### Dropping support for PyPDF2
I don't use PyPDF2 either. The last time I tried was four year ago and I could not get it to work either. But I just checked and the library seems to have a new version that is actively developed: https://github.com/py-pdf/pypdf, so perhaps before considering to drop support for it, we should upgrade to the latest version and try it. I think this work should be done as part of a separate PR though.

PR #2495, opened 2023-12-03
Improve support when specifying a pre-existing toolbar as the target of agw.aui Minimize

jmoraleda:
agw.aui includes the option to specify a pre-existing toolbar as the location where the _restore icon/tool_ should go when minimizing panes -- instead of the default behavior which is creating a new automatic toolbar.

But the option of specifying a pre-existing toolbar cannot handle loading new perspectives or closing or detaching minimized panes. The main issues when specifying a pre-existing toolbar are:

* Using the tool id to find the tool does not work well because multiple tools with the "ID_RESTORE_FRAME" id could have been added to the same toolbar
* Using the pane label/caption to find the tool does not work well when loading a perspective because the caption of the pane could be changed by the load operation
* When loading a perspective, a pane that was minimized could become unminimized or vice versa, and when using a pre-existent toolbar, the toolbar would be preserved by the perspective change operation and the restore icon would need to be added or deleted from the toolbar as appropriate, but not added twice if it was already there.

This PR addresses these problems.


jmoraleda:
This is a new-feature PR as opposed to a bug-fix PR. It does not impact existing code in any way.

I and others have been testing it for several months as part of the wxPythonMeticy release and it appears to work well allowing the user to specify a pre-existing toolbar as the target to place the "restore" icon when minimizing an icon, so the end user has control or where the restore icons will appear.

This functionality in turn requires some additional data to be saved when saving a perspective so the perspective can be loaded preserving the target toolbar.

Is there anything I could I do to help review it and approve it?

PR #2500, opened 2023-12-20
wx.agw.aui. Do layout as the last step after all pane infos have recomputed their best sizes

jmoraleda:
In python only aui, the layout of the docks and panes should be the last step of the update, after all panes have recomputed their best sizes.


Metallicow:
Looks good to me on the Optimizing front...

PR #2501, opened 2023-12-21
wx.agw.aui: Make behavior in all platforms more equal

jmoraleda:
This is a very small PR. It removes an unnecessary `if '__WXGTK__' in wx.PlatformInfo:` Removing it makes the code base two lines smaller, as well as make the behavior in different platforms more equal in some corner cases.

Remark that the change is to _defer_ the actual update (DoUpdate) in all platforms, as it is currently done in GTK. This makes sense as `Update` is called internally by various functions inside framemanager and not all resizing of components may have taken place when it is called.

The following images show one such corner case when this deferral makes the different between incorrect and correct layout. In particular when aui panes have callbacks that impact dock sizes when they are added programmatically to a notebook so that the dock becomes taller to make room for the tabs in the notebook. (The patient data shows is *not* from an actual patient.)

### GTK (same, correct, behavior before and after the proposed change)
image

### MSW (incorrect behavior before the change. Notice how the patient data is cut because the dock tabs take space)
image

### MSW (correct behavior after the change)
image

### DIFF
--- a/wx/lib/agw/aui/framemanager.py
+++ b/wx/lib/agw/aui/framemanager.py
@@ -6141,11 +6141,8 @@ class AuiManager(wx.EvtHandler):
         return self._dock_constraint_x, self._dock_constraint_y
 
     def Update(self):
-        if '__WXGTK__' in wx.PlatformInfo:
-            wx.CallAfter(self.DoUpdate)
-        else:
-            self.DoUpdate()
-
+        wx.CallAfter(self.DoUpdate)
+        
     def DoUpdateEvt(self, evt):
         self.Unbind(wx.EVT_WINDOW_CREATE)
         wx.CallAfter(self.DoUpdate)


Metallicow:
While that looks correct, I have taken the stance to fork agw.aui in my project...
... which also led me to numerous fixes and such I have shared with all "you's"
... but in the end it still lead up to the same result for me. RIP agw.aui all apart and rewrite it to work right from the getgo instead of monkey/hotpatching everything.
... This is a tedious type of attack and takes a lot of experience, but if you want to keep trying to fix the damned bird, then by all means do so, as it seems others are still making progress on that front.
... it doesn't mean that other design implementations will not be considered, it's just that you really have to stay on one side of the fence or another. The grass is always greener on the other side so to speak. CPU vs GPU ect...

jmoraleda:
@Metallicow I completely agree.  This is exactly why I closed: https://github.com/wxWidgets/Phoenix/pull/2473 without waiting to be merged. My current feeling is that I want to submit a PR to tackle the race conditions that occur rather under intensive usage in the current implementation of agw.aui, but only after thorough testing suggests that with those patches the errors are materially reduced. 

(This PR where we are comenting on is not that: I was not attempting to fix any bug. Rather I  wanted to make the behavior more equal across platforms which I think is always desirable)

Metallicow:
@jmoraleda I see your direction...
It started just as simple fixing thing and then rRIPPING them out from awg.aui. 
This is because the original aui guy/gal fell off the planet and we are left with this stuff Andrea ported.

I have since ported the whole wxpy demo and whatnot, so it seems easiest to start with a clean slate(call it a movable panel) as basics and work from there(many base widgets are available). Walk thru all unit tests/DesignImplementations one by one and make sure they pass. Don't proceed on a fork unless the base implementation passes. For this type of simple geometry... there should be no exceptions as to how it should work, I understand this is opensource, but take pride in your work to get it ported into the standard library.

Note: Some of us have not our widgets into the standard library yet.... why? Probably because they(TimeBearers/CoreDevelopers)(so to speak don't have TIME(to review or otherwise), or they haven't fleshed out their code that makes it easy to port into c/c++(this takes time also...)) have other things todo in their lives. 

Anyhow.... Keeping a good eye on the current flawed implementation for fixes is still a good idea. Don't get me wrong... that is the reason why I have a custom GUI I written completely myself.

There is the "wxPyDemo" but not everyone uses it before opening or commenting on a issue on github.... haha

Thank you for your work.time.  :~

Metallicow:
By the way, I had done something similar in my mcow.agw private version... I moved the DoUpdate(method) elsewhere....

D2R20:
Hello everyone, 

I just wanted to add a comment on this one. First I would like to state that I dont master the agw.aui and I dont know what's inside the DoUpdate method. I just use the aui_manager in apps I develop.

I want to give a feedback from a wxpython user.

I work on MSW, so in wxpython 4.2.1, calling aui_manager.Update in fact does 'self.DoUpdate()' and from wxpython 4.2.2 calling aui_manager.Update does 'wx.CallAfter(self.DoUpdate)'.

In my case, this modification introduces a bit of flickering. Indeed, with wxpython 4.2.2, when I trigger the Update method, i can see the new pane areas, but with empty panes for a few moment (~0.3 sec in my case) afterwards the panel inside are drawn.

D2R20:
Hello @jmoraleda, thanks for your quick reply. Unfortunately, I tried to make a small scale example, but in that case the flickering could not be seen.

What I did is to put everything after aui_manager.Update into a wx.CallAfter to force the DoUpdate to be done at first. It solved the flickering. So previously it was:

some_pane.Show()
aui_manager.Update()
some_work()

I modified into:

some_pane.Show()
aui_manager.Update()
wx.CallAfter(some_work)

I assume your suggestion would work as well as there's the self.DoUpdate outside the wx.CallAfter as in wxpython==4.2.1

I chose to modify my code so that it is futureproof

PR #2502, opened 2023-12-22
Fix additional SyntaxWarnings with Python 3.12

swt2c:
It seems that Python 3.12 has started emitting additional SyntaxWarnings in cases where there are backslashes used in strings, so change these strings to raw strings to avoid the warnings.

Fixes #NNNN

### DIFF
--- a/wx/lib/docview.py
+++ b/wx/lib/docview.py
@@ -1023,7 +1023,7 @@ class DocTemplate(wx.Object):
 
 
     def __init__(self, manager, description, filter, dir, ext, docTypeName, viewTypeName, docType, viewType, flags=DEFAULT_TEMPLATE_FLAGS, icon=None):
-        """
+        r"""
         Constructor. Create instances dynamically near the start of your
         application after creating a wxDocManager instance, and before doing
         any document or view operations.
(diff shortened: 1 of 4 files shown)


Metallicow:
Yes, this seems to be correct. This type of fixing behaviour has been around a long time for me. Dumb as it is...

a u

a k

a anyChar type string pulls this "shenanigans"

Metallicow:
Stylistic Issue is the correct terminology

Metallicow:
@swt2c Are you OK? LOL
... Being serious here ...

swt2c:
> @swt2c Are you OK? LOL ... Being serious here ...

Why wouldn't I be?

Metallicow:
... Well ... the way "you" wrote the issue and how its spoken and how it was replied to you, kind of all matters, but only in python.
It appears that from your comment that you was not trying to make a joke.

... after all folks from Notepad++ sometimes like to inject their political humors and whatnot into the executable after build point, which is an interesting point, but not good enough to excert a laugh out of me as you have with a seemingly intellectual prowess that denotes nothing but a good black joke.

@swt2c Was you joking? LOL
... Being serious here ...

PR #2528, opened 2024-02-13
speedmeter.py lost background color

decodeais:
Missing typcasts make the background of the speedmeter white
This problem seems to appear since python 3.10 on windows, linux, raspian ... After  this change it works. please look in the other instruments for similar mistakes, please.

### DIFF
--- a/wx/lib/agw/speedmeter.py
+++ b/wx/lib/agw/speedmeter.py
@@ -840,7 +840,7 @@ class SpeedMeter(BufferedWindow):
             interface = 0
 
             for ind in range(numsteps+1):
-                currCol = (r1 + int(rf), g1 +int(gf), b1 +int(bf))
+                currCol = (int(r1 + rf), int(g1 + gf), int(b1 + bf))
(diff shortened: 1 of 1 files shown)


wx/lib/agw/speedmeter.py
```diff
@@ -840,7 +840,7 @@ def Draw(self, dc):
             interface = 0
 
             for ind in range(numsteps+1):
-                currCol = (r1 + rf, g1 + gf, b1 + bf)
+                currCol = (r1 + int(rf), g1 +int(gf), b1 +int(bf))
```
swt2c, line 843:
```suggestion
                currCol = (int(r1 + rf), int(g1 + gf), int(b1 + bf))
```

decodeais:
Should work too. The problem was the floating-point value and your solution
fixes it.

Scott Talbert ***@***.***> schrieb am Di., 13. Feb. 2024,
23:25:

wx/lib/agw/speedmeter.py
decodeais: Its OK

PR #2536, opened 2024-02-27
update BusyCursor snippet

### LINKED ISSUE #2535: BusyCursor docs are outdated
**wxPython version & source**:          4.2.1, pypi

**Description of the problem**: wx.BusyCursor docs should detail the context manager case, which the cursor is actually meant to be used like.

<details>

```python
with wx.BusyCursor():
    for i in range(50000):
        DoACalculation()
```
</details>

DietmarSchwertberger:
Fixes #2535 by adding a context manager example and replacing xrange with range.

### DIFF
--- a/docs/sphinx/rest_substitutions/snippets/python/converted/wx.BusyCursor.1.py
+++ b/docs/sphinx/rest_substitutions/snippets/python/converted/wx.BusyCursor.1.py
@@ -1,7 +1,11 @@
 
+        # Normal usage
         wait = wx.BusyCursor()
-
-        for i in xrange(10000):
+        for i in range(10000):
             DoACalculation()
-
         del wait
+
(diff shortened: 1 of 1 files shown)


swt2c:
Thanks Dietmar!

Metallicow:
I guess i didn't search over the sphinx generated docs for xrange after everything was commited. Shame on me. Good catch.
... Probably might be some more py2 to py3 changes lingering if you found that. I never uploaded a list of what [py2 to py3]I had done, cause back then it was constantly phoenix changing and havok. It might be better now to make a list and thoroughly check it considering phoenix has a "quote" stable version now (wxPython 4.0.7.post2 msw (phoenix) wxWidgets 3.0.5 Scintilla 3.21).
... more or less ...

PR #2540, opened 2024-03-20
remove Python2 hybridation code

a-detiste:
`print()` is available since Python 2.6

https://wiki.debian.org/Python3-six-removal


samples/printing/printing.py
```diff
@@ -1,5 +1,6 @@
+from __future__ import print_function
```
swt2c, line 1: Probably don't need this...we can drop support for Python 2.

a-detiste: so let's remove all `six`usage ?

a-detiste: ... and the vendored copy living as `wx/lib/pubsub/py2and3.py`

wx/lib/agw/ultimatelistctrl.py
```diff
@@ -550,7 +549,7 @@ def __init__(self, parent):
 IL_VARIABLE_SIZE = 1
 
 # Python integers, to make long types to work with CreateListItem
-INTEGER_TYPES = six.integer_types
+INTEGER_TYPES = int
```
bewczardski, line 552: This is changing the value from a list to a scalar value which is now causing crashes later on in the source. I converted all `in INTEGER_TYPES` to `== INTEGER_TYPES` and it fixed it. Should probably update the name to reflect the change as well.

`if type(itemOrId) in INTEGER_TYPES:` < line 598 

Another example:
```class CommandListEvent(wx.PyCommandEvent):
    """
    A list event holds information about events associated with :class:`UltimateListCtrl`
    objects.
    """

    def __init__(self, commandTypeOrEvent=None, winid=0):
        """
        Default class constructor.
        For internal use: do not call it in your code!

        :param `commandTypeOrEvent`: the event type or another instance of
         :class:`PyCommandEvent`;
        :param `winid`: the event identifier.
        """

        if type(commandTypeOrEvent) in INTEGER_TYPES:
````

**Comment by a-detiste** [comment]
Thanks for the review

The right thing to do is to get rid of INTEGER_TYPES altogether:

```
-if type(commandTypeOrEvent) in INTEGER_TYPES:
+if isintance(commandTypeOrEvent, int):
```

jmoraleda:
I agree with this PR. I also agree with removing all `six` usage.

komoto48g:
I agree with dropping support for Python 2.
IMHO, this PR is too aggressive and the first commit (removing print_) would be more acceptable if it was a separate PR.
> From "https://wxpython.org/pages/contributor-guide/", 
> 5. Do not put more than one fix/feature in the same branch or PR. Make a new distinct branch and PR for each of them.

a-detiste:
I m on Holliday. Anyone can cherrypick the first commit that removes py2.6< compatibility if it tastes them, 
I ll redo this work in smaller pieces when I get back.

a-detiste:
At first there was only this one first commit untill I got asked to proceed further

newville:
Please merge.

a-detiste:
The PR against GRASS (which uses Wx by the way) had a similar +/- size and went in smoothly. I m not scared. What do we need to go further ?

swt2c:
Replaced by https://github.com/wxWidgets/Phoenix/pull/2688

PR #2552, opened 2024-05-17
Update wxWidgets to v3.2.5.

jmoraleda:
Update wxWidgets to v3.2.5. Add support for new API MSW-only Accessibility related methods in wxWidgets. See [relevant wxWidgets commit](https://github.com/wxWidgets/wxWidgets/commit/81e9373efa99aede6a9cd6d280b5671fe8e44be3).

### DIFF
--- a/etg/window.py
+++ b/etg/window.py
@@ -251,6 +251,22 @@ def run():
             return NULL;
         #endif
         """)
+    c.find('GetOrCreateAccessible').setCppCode("""\
+        #if wxUSE_ACCESSIBILITY
+            return self->GetOrCreateAccessible();
+        #else
+            wxPyRaiseNotImplemented();
+            return NULL;
+        #endif
+        """)
+    c.find('CreateAccessible').setCppCode("""\
+        #if wxUSE_ACCESSIBILITY
+            return self->CreateAccessible();
+        #else
+            wxPyRaiseNotImplemented();
+            return NULL;
+        #endif
+        """)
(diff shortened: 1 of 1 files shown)


swt2c:
Thanks, was going to do this myself soon.  Let's see how the CI does...

swt2c:
Actually yeah I think we probably want Dietmar's version of the accessibility stuff.  So let's drop that from this one and just update to wx 3.2.5 (or I can do that).

jmoraleda:
What about we do it in two phases? This PR, does the minimal update to 3.2.5 (without the minimal `windows.h` stuff that I have here it won't compile on GTK or Mac). And then we can do the the other PR to actually add Accessibility support. What do you think?

swt2c:
Can you update this to include Dietmar's changes?  If not, I'll do it.

jmoraleda:
Scott, thank you for offering to help. Please, go ahead and make the changes. Either as part of this PR, or merge this PR and then immediately add Dietmars'. As you think best.

I am not comfortable enough with sip to make Dietmar's changes: I was able to make the changes here because the new functions had identical signatures to `GetAccessible` which was already handled in sip.

swt2c:
You don't have to know anything about sip, just how to copy and paste.  Anyway, I merged your changes with Dietmar's and pushed them (I couldn't push to your branch even though github says I should be able to edit your PR).

jmoraleda:
Thank you, Scott.

PR #2561, opened 2024-06-13
Add Wayland support for Window.GetHandle

### LINKED ISSUE #1217: GetHandle() returns 0 on Linux
**Operating system**: Manjaro Linux x64 (updated today), GTK 3.24.8-1, GNOME 3.32.1
**wxPython version & source**:        wxPython 4.0.4 (pip) (also tested on wxPython 4.0.2 [venv pip])
**Python version & source**:            Python 3.7.3 (Stock from the Manjaro-Packages, and venv, booth having the same problem)

I was trying to run the CEFPython-wxPython example app from [https://github.com/cztomczak/cefpython/blob/master/examples/wxpython.py](https://github.com/cztomczak/cefpython/blob/master/examples/wxpython.py). While it did run fine on my Windows machine it always stopped at line 174 `assert self.browser_panel.GetHandle(), "Window handle not available"` because the function `GetHandle()` returned an integer 0 on Linux.
I did a little debugging and found out, that even if I wait up to 8 seconds I do not get an Handle for the Frame or for the Panel. Did a little Debugging on another application of mine and found out that `GetHandle()` always returns an integer 0 there for every Frame and Panel available.

unawarez:
Fixes #1217

This PR adds support for getting a `wl_surface` from Window.GetHandle when running on GTK+Wayland. This patch was originally more complicated by trying to support all the GTK windowing backends. Now, any GTK backend besides X11 or Wayland will return 0. This is less confusing than trying to figure out and match the return type of C++ GetHandle on Windows/Mac/etc in the unlikely case someone uses the GTK port on those platforms. This maintains backwards compatibility because up to now, GetHandle failed and returned 0 on any non-X11 GTK backend.

I tested on Wayland and X11 on current Arch with Python 3.12. Both work (at least, return a nonzero number) as long as you call `frame.Show()` before calling `frame.GetHandle()`. Also confirmed ARM Mac without GTK hasn't broken. I was not able to test Windows.

Suggestions welcome!

### DIFF
--- a/etg/window.py
+++ b/etg/window.py
@@ -136,6 +136,16 @@ def run():
     m1.find('externalLeading').out = True
 
     c.find('GetHandle').type = 'wxUIntPtr*'
+    c.find('GetHandle').detailedDoc = [
+        """The returned value differs from the C++ version of GetHandle when \
+        running on the GTK port. When running on Wayland with GTK, this \
+        function returns a `wl_surface` pointer for the native OS window \
+        containing the widget. On X11 with GTK, this returns the X window \
+        ID for the containing window. On any other backend with GTK, this \
+        function returns 0.\n\n"""
+
+        """On some platforms this may return 0 if the window has not yet been shown."""
+    ]
     c.find('GetHandle').setCppCode("return new wxUIntPtr(wxPyGetWinHandle(self));")
 
     c.addCppMethod('void*', 'GetGtkWidget', '()', """\
--- a/src/window_ex.cpp
+++ b/src/window_ex.cpp
@@ -1,32 +1,15 @@
-
-#ifdef __WXMSW__
-#include <wx/msw/private.h>
-#endif
-
 #ifdef __WXGTK__
-#include <gdk/gdkx.h>
+#include <gdk/gdk.h>
 #include <gtk/gtk.h>
-#ifdef __WXGTK3__
-// Unlike GDK_WINDOW_XWINDOW, GDK_WINDOW_XID can't handle a NULL, so check 1st
-static XID GetXWindow(const wxWindow* wxwin) {
-    if ((wxwin)->m_wxwindow) {
-        if (gtk_widget_get_window((wxwin)->m_wxwindow))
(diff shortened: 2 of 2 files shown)


etg/window.py
```diff
@@ -136,6 +136,16 @@ def run():
     m1.find('externalLeading').out = True
 
     c.find('GetHandle').type = 'wxUIntPtr*'
+    c.find('GetHandle').detailedDoc = [
```
unawarez, line 139: I don't know what the right API to do this is, and I couldn't get inline markup working right.

etg/window.py
```diff
@@ -136,6 +136,16 @@ def run():
...
+        """The returned value is the platform-defined handle for the native \
+        window containing the widget. On Windows this is an HWND, on X11 it \
+        is the X window ID, on Wayland it is a wl_surface, etc.\n\n"""
```
unawarez, line 142: The weasely "etc" is because I'm not sure what wxWidgets on Cocoa returns. My added code for GTK on Mac returns NSWindow, but I just spent a long time in the wxWidgets source and can't figure out what GetHandle will return. The good thing is at least for Cocoa this PR doesn't affect the return value so existing code should still work the same.

unawarez:
Moved back to draft because of additional Mac stuff.

I mentioned I wasn't sure what wxWidgets GetHandle actually returns on non-GTK Mac. Well, this quick test suggests it's something widget-like as opposed to the window handle:

```
>>> import wx
>>> app=wx.App()
>>> frame=wx.Frame(None)
>>> child=wx.StaticText(frame)
>>> frame.Show()
True
>>> frame.Handle
4967401760
>>> child.Handle
4967403792
```

If upstream GetHandle was returning NSWindow, those two handles should be the same.

So:
* It seems bad to me if I differ too strongly here, by returning the root window for any widget instead of returning the widget.
* But, this only affects Mac+GTK builds, which a) probably no one uses, and b) if someone did, the current GetHandle was always broken for them because it's hardcoded to getting an X11 window. So maybe this doesn't really matter?
* I don't know anything about Cocoa/ObjC/...
* Could really use an opinion about what to do here!

unawarez:
Radically simplified & should be ready to go.

PR #2566, opened 2024-06-21
Add Window methods GetNativeWindowHandle and GetNativeWindowHandleType

unawarez:
This adds 2 new Window methods that give users access to the platform window handle. C++ `GetHandle` usually returns widget handles. The user can technically get from there to the window handle on their own, but it's very complicated and would require guessing some info that wxWidgets already knows.

My use-case is creating an OS child window in order to embed Vulkan graphics in a wxPython GUI. It also looks like folks use this trick for embedding webviews.

`GetNativeWindowHandle` returns a pointer-as-int, and `GetNativeWindowHandleType` returns a string designating what the real type of `GetNativeWindowHandle()` is. Both these functions especially help GTK users by "dereferencing" down to the actual MSW/Mac/X/Wayland backend that GTK is using.

This includes a new unit test that tries to confirm on any platform that:
* `NativeWindowHandle` is not zero
* For a parent and child widget in the same window, `parent.NativeWindowHandle == child.NativeWindowHandle`.
This seems like a pretty solid check, without any platform knowledge, that whatever is being returned really is a window handle and not a widget handle.

Returning strings from `NativeWindowHandleType` isn't great. I'm not sure what the best way to go is.

Tested on:
- [X] Mac
- [X] Linux GTK X11
- [X] Linux GTK Wayland
- [ ] Windows

### DIFF
--- a/etg/window.py
+++ b/etg/window.py
@@ -240,8 +240,8 @@ def run():
             return pix
             """)
 
-    c.addCppMethod('void*', 'GetNativeWindowHandle', '()',
-           'return wxPyNativeWindowHandle(self);',
+    c.addCppMethod('size_t', 'GetNativeWindowHandle', '()',
+           'return (size_t)wxPyNativeWindowHandle(self);',
            doc="""Returns a platform-specific handle to the OS window containing this widget.
 
                 The :meth:`GetNativeWindowHandleType` method provides the concrete type of the handle.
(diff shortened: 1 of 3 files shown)


unawarez:
I'm no longer sure this helps the use-cases ☹️ it's still very hard to create a second window and have its event processing work in any sane way, and adding more wx hooks like this isn't going to solve that.

unawarez:
Postmortem: getting the toplevel window is counterproductive for rendering on Windows/Mac/X anyway.

Issue #2498, opened 2023-12-16
FloatCanvas triggers DeprecationWarning from Numpy 1.25.0 or later for 'sometrue'


reticulatus:
**Operating system**: Linux Mint 21.2
**wxPython version & source**:  4.2.1  from pypi         
**Python version & source**: 3.10.12 from distro          
**Numpy version & source**: 1.26.2 from pypi

**Description of the problem**:

I copied the code from the [docs for the FloatCanvas package](https://docs.wxpython.org/wx.lib.floatcanvas.FloatCanvas.html#module-wx.lib.floatcanvas.FloatCanvas) into a small test app (see example below). 

When I ran the app, it output the following warning:
```
/home/richardt/Development/python/exp/hci/wxpython/float_canvas/float_canvas_1.py:25: DeprecationWarning: `sometrue` is deprecated as of NumPy 1.25.0, and will be removed in NumPy 2.0. Please use `any` instead.
  self.Canvas.Draw()
```

Line 604 in FloatCanvas.py for the FloatCanvas.Draw() method contains:
```
        if N.sometrue(self.PanelSize <= 2 ):
```

I did a recursive search of my wxPython 4.2.1 installation and found that `sometrue` also occurs in line 2047 of  `lib/plot/plotcanvas.py`:
```
             if np.sometrue(
```

<details>
  <summary>Code Example (click to expand)</summary>

```python
import wx
from wx.lib.floatcanvas import FloatCanvas

class MyFrame(wx.Frame):
    def __init__(self, parent):
        wx.Frame.__init__(self, parent)

        self.Canvas = FloatCanvas.FloatCanvas(self, -1,
                                     size=(500, 500),
                                     ProjectionFun=None,
                                     Debug=0,
                                     BackgroundColor="White",
                                     )

        # add a circle
        cir = FloatCanvas.Circle((10, 10), 100)
        self.Canvas.AddObject(cir)

        # add a rectangle
        rect = FloatCanvas.Rectangle((110, 10), (100, 100), FillColor='Red')
        self.Canvas.AddObject(rect)

        self.Canvas.Draw()

app = wx.App()
frame = MyFrame(None)
frame.Show()
app.MainLoop()

```
</details>

Issue #2574, opened 2024-07-11
Numpy2 removed numpy.sometrue. 


iandobbie:
Windows 10 22H2
'4.2.1 msw (phoenix) wxWidgets 3.2.2.1'
Python 3.10.7 (tags/v3.10.7:6cc6b13, Sep  5 2022, 14:08:36) [MSC v.1933 64 bit (AMD64)] on win32

I am not yet sure exactly the code path that is calling this but I will track down how to trigger this in a simple way but my application generates an exception. The core issue is in floatcanvas/FloatCanvas.py at line 604 

        if N.sometrue(self.PanelSize <= 2 ):

numpy no longer has a numpy.sometrue() function

iandobbie:
If you just try to run Phoenix/samples/floatcanvas/MicroDemo.py you actually get two different numpy2 errors. 

  File ".../floatcanvas/lib/python3.12/site-packages/wx/lib/floatcanvas/FloatCanvas.py", line 782, in ZoomToBB
    (BoundingBox[0,1]+BoundingBox[1,1])/2 ),N.float_)

Numpy.float_ has been removed from numpy2

and then 

  File ".../floatcanvas/lib/python3.12/site-packages/wx/lib/floatcanvas/FloatCanvas.py", line 604, in Draw
    if N.sometrue(self.PanelSize <= 2 ):
       ^^^^^^^^^^

numpy.sometrue has also been removed. 

Changing these sections to N.float64 and N.any respectively fixed these issues.

reticulatus:
Yes, `numpy.sometrue()` has been raising DeprecationWarnings for a while now, see https://github.com/wxWidgets/Phoenix/issues/2498

PR #2578, opened 2024-07-15
Replace use of numpy.sometrue with numpy.any for numpy v2 compatibility (#2498)

carandraug:
Numpy v2 has removed the `numpy.sometrue` function (it was deprecated in numpy 1.25.0). numpy.any has been around for a very long time so there should be no backwards compatibility issue.

This fixes/closes issues #2498 and #2574.


iandobbie:
The MicroDemo.py also showed up an issue with numpy.float_ see #2574. So maybe this patch should also fix that issue to solve more numpy2 compatibility problems.

PR #2591, opened 2024-08-29
Replace N.alltrue() with N.all()

s-t-e-v-e-n-k:
Building on the excellent work in f7d1d818, replace the two uses of N.alltrue() with N.all(), since the former has been removed in numpy 2.0.


bnavigator:
`float_` must also be removed.

swt2c:
@s-t-e-v-e-n-k can you also eliminate `float_` as @bnavigator suggested?

s-t-e-v-e-n-k:
> @s-t-e-v-e-n-k can you also eliminate `float_` as @bnavigator suggested?

Done!

PR #2580, opened 2024-07-19
Fix NumPy 2.0 deprecations via running ruff

radoering:
I just run the following command:

```
ruff check --select NPY201 --fix --exclude docs/sphinx/rest_substitutions/snippets/python/converted
```

to fix numpy deprecations that were removed and thereby result in an exception with numpy 2.0.

See https://numpy.org/devdocs/numpy_2_0_migration_guide.html#numpy-2-0-migration-guide

Fixes #2574
Fixes #2498
Closes #2578

PR #2592, opened 2024-08-29
replace old numpy types no longer valid in numpy 2

bnavigator:
Follow up on #2580. Ruff did not catch this one because the nonstandard `N` for numpy is only imported  via `from wx.lib.floatcanvas.Utilities.BBox import *`


bnavigator:
(Partial) duplicate of #2591

PR #2600, opened 2024-09-05
Possible fix for #2345

### LINKED ISSUE #2345: Thumbnailctrl SetSelection raises exception if it tries to scroll
**Operating system**:          Windows 10
**wxPython version & source**:          4.2.0 msw (phoenix) wxWidgets 3.2.0
**Python version & source**:            Python 3.10.6

<details>
  <summary>Thumbnailctrl SetSelection raises exception if it tries to scroll</summary>

When calling SetSelection, if the thumbnail control tries to scroll the selected thumbnail into view, it will fail with the error:
```
  File "E:\wxpython\wxPython-demo-4.1.1\demo\agw\ThumbDemoConfig.py", line 416, in OnSelectFirst
    self.scroll.SetSelection(0)
  File "C:\Program Files\Python310\lib\site-packages\wx\lib\agw\scrolledthumbnail.py", line 1269, in SetSelection
    self.ScrollToSelected()
  File "C:\Program Files\Python310\lib\site-packages\wx\lib\agw\scrolledthumbnail.py", line 1540, in ScrollToSelected
    self.Scroll(x,sy)
TypeError: _ScrolledWindowBase.Scroll(): arguments did not match any overloaded call:
  overload 2: argument 1 has unexpected type 'int'
```

wx/lib/agw/scrolledthumbnail.py - line 1537. This is using float division. I believe the only change needed is to switch to // instead of /. There may be other instances of float division where int division was intended.

This can be reproduced with a small change to the thumbnail demo, specifically ThumbDemoConfig.py:

```diff
86a87
>         self.selectfirst = wx.Button(self.panel, -1, "Select first")
110a112
>         self.Bind(wx.EVT_BUTTON, self.OnSelectFirst, self.selectfirst)
168a171
>         optionsizer.Add(self.selectfirst, 0, wx.TOP|wx.LEFT|wx.ADJUST_MINSIZE, 3)
411a415,416
>     def OnSelectFirst(self, event):
>         self.scroll.SetSelection(0)
```

This just adds a button that will select the first item. If the current directory has enough images to show a scrollbar and you scroll down before clicking the button, it will raise the exception.

</details>

infinity77:
Possible fix for #2345 - Thumbnailctrl SetSelection raises exception if it tries to scroll

Fixes #2345

### DIFF
--- a/wx/lib/agw/scrolledthumbnail.py
+++ b/wx/lib/agw/scrolledthumbnail.py
@@ -1534,7 +1534,7 @@ class ScrolledThumbnail(wx.ScrolledWindow):
 
         # scroll view
         xu, yu = self.GetScrollPixelsPerUnit()
-        sy = sy/yu + (sy%yu and [1] or [0])[0] # convert sy to scroll units
+        sy = sy//yu + (sy%yu and [1] or [0])[0] # convert sy to scroll units
         x, y = self.GetViewStart()
 
         self.Scroll(x,sy)


swt2c:
@infinity77 thanks for the fixes!

PR #2609, opened 2024-09-13
pdfviewer: Add support for pymupdf renaming

Infernio:
The 'fitz' module got renamed to 'pymupdf' in 1.24.3. Prepare for an eventual deprecation and removal by importing from pymupdf first and falling back to fitz only if that fails.

I also got rid of the other 'fitz'es in the file via 'fitz as pymupdf', since pymupdf.open is much easier to read than fitz.open.


swt2c:
Thanks!

PR #2610, opened 2024-09-14
fix siplib build on Windows

DietmarSchwertberger:
When trying to build 4.2.2 with Python 3.12 on Windows 64 bit, I got an error when siplib was linked.
The symbol `sipSetBool` was missing.

The generated `sip\siplib\setup.py` has these lines, but the file content is not used when building with waf:
```
if sys.platform == 'win32':
    module_src.append('bool.cpp')
```

This PR adds `bool.cpp` also to `wscript`.

### DIFF
--- a/wscript
+++ b/wscript
@@ -9,6 +9,7 @@
 
 import sys
 import os
+import glob
 import setuptools
 
 try:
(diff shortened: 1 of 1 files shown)


swt2c:
bool.cpp was removed in sip 6.8.5.  Is there some reason you can't use sip 6.8.5?

DietmarSchwertberger:
Thanks. This was not intentional. I thought build.py would check for correct prerequisites?

swt2c:
> Thanks. This was not intentional. I thought build.py would check for correct prerequisites?

It does for some things that it depends on (like doxygen and waf) but not for python dependencies, unfortunately as those kind of need to be installed outside of its purview.  We could probably do something like that if we add a pyproject.toml, but that would mostly work for the <build from sdist case> but I'm not sure about using build.py directly.

PR #2617, opened 2024-10-01
Fix problems identified by dependabot

HelioGuilherme66:
Changes versions of modules with vulnerabilities issues identified by `dependabot`on my fork.


swt2c:
I'm thinking probably we'll hold off on this for especially for Jinja2 and markupsafe.  Those are both used for docs and I had a hard time finding compatible versions when I built docs for 4.2.2, so I'd rather leave it alone and get rid of those pins when I get a newer sphinx working correctly.

swt2c:
This has been resolved by #2695.

PR #2468, opened 2023-10-18
Improve Python type-stubs

lojack5:
The wxPython type-stubs are nice in that they expose what classes, methods, etc are available, but there is no type information in them (there's even a comment in `makePyArgsString` referencing implementing this).  This is my 90% of the way there implementation of exposing this type information to the type-stubs.

I also didn't find any open issues regarding type-stubs, so I haven't linked an issue.

For the most part, all of the type-information was already there in the doxygen files, it just had to be pulled into the args strings. There are a few exceptions, mostly revolving around types that are C++ typedefs (ex: WindowID and Coord).

By using `from __future__ import annotations`, all of the type-hints are stringized, so even in the cases where the name referenced is undefined (ex: richtext's TextAttrDimensionFlags), the type-stub files are still valid.  The hints in those cases will just provide no information.

I do have a few discussion points to probably resolve before this is truly ready though:
- From what I can tell by the buildbot output, we build as low as Python 3.6. So I'm assuming any new code doesn't need to be compatible with earlier versions, is this a good assumption?
- How to handle `typing.ParamSpec`: related to the above, ParamSpec was added in Python 3.10. Currently it's only used to type-hint `CallAfter` and `CallLater`, how would you like this import handled:
  - If you're OK with a new dependency [typing-extensions](https://pypi.org/project/typing-extensions/) could be added for Python versions < 3.10, then import from there.
  - I can skip out on typing CallAfter and CallLater, although if other methods come up in the future that would benifit from this sort of typing, we run into the issue again.
- Code style - I tried to keep with the same code style in surrouding areas but I probably named a few variables differently at some point. I also used f-strings since those are in Python 3.6+.  I can go back and use Python 2 style formatting strings if that's what's wanted however.
- There are still some typedef types that aren't covered.  I could dig through and find every instance in the PYI files where a type is referenced that's undefined, but that may end up with a lot of hard-coded conversions (usually to int) in the type-name fixup code. Let me know of any changes there I need to make.
- I have no knowledge of PI files (WingIDE files according to the header), do these changes break those?
[...]

PR #2668, opened 2025-01-12
Type stub improvements

### LINKED ISSUE #2665: Python type-stubs problems
**wxPython version & source**: [wxPython-4.2.3a1.dev5812+c3092be4-cp311-cp311-win_amd64](https://github.com/wxWidgets/Phoenix/actions/runs/12644504611)
**Python version & source**: Python 3.11.6

Thanks a lot @lojack5 for creating #2468 and @swt2c for merging it. I tried the above mentioned snapshot in my codebase and found the following problems with VSCode and pylance v2024.12.1:

1. In `wx.Frame`, `wx.Dialog` and every class that inherits from it, the parameter `parent` can also be `None` and not only `wx.Window`.
image

2. Properties are not correctly typed. Eg. `wx.Size().GetWidth()` is `int`, but `wx.Size().Width` is of type `Any`.
image

3. `wx.TIMER_CONTINUOUS` & `wx.TIMER_ONE_SHOT` should be `bool` instead of `int`, because in `Timer.Start(...)` the parameter `oneShot` is also `bool`
image

4. `wx.TreeItemId.__bool__()` has to return `bool` instead of `int`
image

5. `wx.dataview.DataViewItem.__bool__()` has to return `bool` instead of `int`
image

6. `wx.DateTime` can be constructed from `datetime.datetime` and `datetime.time`, so an `__init__` overload should be available for these types.
image

### LINKED ISSUE #2666: Warnings when building documentation after type stubs changes
After the type stubs changes, when running `build.py etg` (without --no-doc), the process emits a bunch of warnings like the below:

```
SEVERE: Incompatibility between function/method signature and list of parameters in `wx.glcanvas.GLAttribsBase.AddAttribBits`:

The parameter `searchVal : int` appears in the method signature but could not be found in the parameter list.

  ==> Function/Method signature from `extractors`: AddAttribBits(searchVal : int, combineVal : int)
  ==> Parameter list from wxWidgets XML items:     ['searchVal', 'combineVal']

This may be a documentation bug in wxWidgets or a side-effect of removing the `wx` prefix from signatures.
```

@lojack5 any change you can address this one?

lojack5:
Follow up to #2468 , addressing issues brought up in #2665 and #2666 . (I finally figured out my problem with running `etg` without `--nodoc`, so now I can finally test the doc building warnings).

Issues addressed:
 - [x] sphinx generator "SEVERE" warnings during signature validation:
   - Over-aggressive C++ syntax removal (ex: `constraint` -> `raint`) no longer happens.
   - Special case handle parameters named `_from`, `_is`, and `_def` (Python reserved keywords)
   - Most other warnings are handled by correctly factoring in (and removing) that the type-hints are present in the args string.
 - [x] `__bool__` returning `int`: Special case handling done in the writing of the method.  I'm not sure why we can't just change the definition in the sip file, but sip complains that it must return an `int`.
 - [x] `None` allowed for `parent` argument on `TopLevelWindow` subclasses.
 - [x] Marks values coming from `#define` directives which are boolean as `bool`.
 - [x] Properties declared using the `variable = property(<getter>, <setter>)` syntax not being seen as the applicable type.  This is a `typeshed` issue (and not one easily solve).  Addressed by using decorator syntax when a getter and type-hints are present on the getter and/or setter (setter-only properties still have to use the `variable = property(fset=<setter>)` syntax.
 - [x] sphinx generator warning for `SetAssertMode`.  The clean-name code was inappropriately removing applying the "remove wx prefix" code to argument names:
  ````
  ==> Function/Method signature from `extractors`: SetAssertMode(AppAssertMode : AppAssertMode)
  ==> Parameter list from wxWidgets XML items:     ['wxAppAssertMode']
  ````
- [ ] `wx.DateTime` accepts `datetime.datetime` and `datetime.date` objects in its constructor.  Similarly, `wx.Point`, `wx.Size`, and other similar automated conversions are handled.

Things I'd like extra eyes on:
- [x] `None` for `parent` argument: Are any of the changed type-stubs incorrect in allowing `None` there? Are there any missing?
- [x] Help with potentially addressing the `wx.DateTime` typing.

Closes #2665 
Closes #2666

### DIFF
--- a/etgtools/extractors.py
+++ b/etgtools/extractors.py
@@ -542,6 +542,10 @@ class FunctionDef(BaseDef, FixWxPrefix):
                     params.append(s)
 
         self.pyArgsString = f"({', '.join(params)})"
+        # __bool__ and __nonzero__ need to be defined as returning int for SIP, but for Python
+        # __bool__ is required to return a bool:
+        if (self.name or self.pyName) in ('__bool__', '__nonzero__'):
+            returns = ['bool']
         if not returns:
             self.pyArgsString = f'{self.pyArgsString} -> None'
         elif len(returns) == 1:
--- a/etgtools/extractors.py
+++ b/etgtools/extractors.py
@@ -19,9 +19,9 @@ from typing import Optional
 import xml.etree.ElementTree as et
 import copy
 
-from .tweaker_tools import FixWxPrefix, MethodType, magicMethods, \
+from .tweaker_tools import AutoConversionInfo, FixWxPrefix, MethodType, magicMethods, \
                            guessTypeInt, guessTypeFloat, guessTypeStr, \
-                           textfile_open, Signature
+                           textfile_open, Signature, removeWxPrefix
 from sphinxtools.utilities import findDescendants
 
 if sys.version_info >= (3, 11):
@@ -506,7 +506,7 @@ class FunctionDef(BaseDef, FixWxPrefix):
                 # now grab just the last word, it should be the variable name
                 # The rest will be the type information
                 arg_type, arg = arg.rsplit(None, 1)
-                arg, arg_type = self.parseNameAndType(arg, arg_type)
+                arg, arg_type = self.parseNameAndType(arg, arg_type, True)
                 params.append(P(arg, arg_type, default))
                 if default == 'None':
                     params[-1].make_optional()
@@ -517,7 +517,7 @@ class FunctionDef(BaseDef, FixWxPrefix):
                     continue
                 if param.arraySize:
                     continue
-                s, param_type = self.parseNameAndType(param.pyName or param.name, param.type)
+                s, param_type = self.parseNameAndType(param.pyName or param.name, param.type, not param.out)
                 if param.out:
                     if param_type:
                         returns.append(param_type)
@@ -696,7 +696,7 @@ class ClassDef(BaseDef):
         self.headerCode = []
         self.cppCode = []
         self.convertToPyObject = None
-        self.convertFromPyObject = None
+        self._convertFromPyObject = None
         self.allowNone = False      # Allow the convertFrom code to handle None too.
         self.instanceCode = None    # Code to be used to create new instances of this class
         self.innerclasses = []
@@ -715,6 +715,18 @@ class ClassDef(BaseDef):
         self.__dict__.update(kw)
         if element is not None:
             self.extract(element)
+
+    @property
+    def convertFromPyObject(self) -> Optional[str]:
+        return self._convertFromPyObject
+    
+    @convertFromPyObject.setter
+    def convertFromPyObject(self, value: AutoConversionInfo) -> None:
+        self._convertFromPyObject = value.code
+        name = self.name or self.pyName
+        name = removeWxPrefix(name)
+        print('Registering:', name, value.convertables)
+        FixWxPrefix.register_autoconversion(name, value.convertables)
 
     def is_top_level(self) -> bool:
         """Check if this class is a subclass of wx.TopLevelWindow"""
--- a/etgtools/extractors.py
+++ b/etgtools/extractors.py
@@ -15,12 +15,13 @@ wxWidgets API info which we need from them.
 import sys
 import os
 import pprint
-import xml.etree.ElementTree as ET
+from typing import Optional
+import xml.etree.ElementTree as et
 import copy
 
-from .tweaker_tools import FixWxPrefix, magicMethods, \
+from .tweaker_tools import FixWxPrefix, MethodType, magicMethods, \
                            guessTypeInt, guessTypeFloat, guessTypeStr, \
-                           textfile_open
+                           textfile_open, Signature
 from sphinxtools.utilities import findDescendants
 
 if sys.version_info >= (3, 11):
@@ -285,6 +286,7 @@ class FunctionDef(BaseDef, FixWxPrefix):
         self.type = None
         self.definition = ''
         self.argsString = ''
+        self.signature: Optional[Signature] = None
         self.pyArgsString = ''
         self.isOverloaded = False
         self.overloads = []
@@ -406,7 +408,10 @@ class FunctionDef(BaseDef, FixWxPrefix):
         else:
             parent = self.klass
         item = self.findOverload(matchText)
+        assert item is not None
         item.pyName = newName
+        if item.signature:
+            item.signature.method_name = newName
         item.__dict__.update(kw)
 
         if item is self and not self.hasOverloads():
@@ -470,8 +475,8 @@ class FunctionDef(BaseDef, FixWxPrefix):
         Create a pythonized version of the argsString in function and method
         items that can be used as part of the docstring.
         """
-        params = list()
-        returns = list()
+        params: list[Signature.Parameter] = []
+        returns: list[str] = []
         if self.type and self.type != 'void':
             returns.append(self.cleanType(self.type))
 
@@ -483,6 +488,7 @@ class FunctionDef(BaseDef, FixWxPrefix):
                         'wxArrayInt()' : '[]',
                         'wxEmptyString':  "''", # Makes signatures much shorter
                         }
+        P = Signature.Parameter
         if isinstance(self, CppMethodDef):
             # rip apart the argsString instead of using the (empty) list of parameters
             lastP = self.argsString.rfind(')')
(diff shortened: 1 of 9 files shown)


swt2c:
Thanks a lot @lojack5 this looks good.

As far as the `wx.DateTime` issue goes, I think it's basically anywhere where a `wx.DateTime` can be passed as a parameter to the C++ side, a `datetime.datetime` or `datetime.date` can also be passed.  Does that help with figuring out a place to do it?

lojack5:
Maybe.

Does the conversion added happen *anywhere* a `wx.DateTime` is accepted as an input to *any* method, or only `wx.DateTime` method?  For example does the conversion happen on `wx.adv.CalendarCtrl.__init__`'s `date` parameter? If so I might be able to do something with that.  If not, then it's a lot more special casing that it might be pretty hard to get right.

If so, I could probably do something like:
```python
DateTimeT: TypeAlias = Union[wx.DateTime, datetime.datetime, datetime.date]
```
Then `wx.DateTime` -> `DateTimeT` for input argument annotations only (ie, be careful not to do this replacement on return type annotations, etc).

Follow on, should something similar be done for `wx.Point`, `wx.Size`, and `wx.Colour`?. Like:
```python
PointT: TypeAlias = Union[wx.Point, tuple[int, int]]
SizeT: TypeAlias = Union[wx.Size, tuple[int, int]]
ColourT: TypeAlias = Union[wx.Colour, tuple[int, int, int], tuple[int, int, int, int], str]
```

etgtools/extractors.py
```diff
@@ -533,10 +533,14 @@ def makePyArgsString(self):
                         default = '|'.join([self.cleanName(x, True) for x in default.split('|')])
                         s = f'{s}={default}'
                     elif param_type:
-                        s = f'{s} : {param_type}'
+                        s = f'{s}: {param_type}'
                     params.append(s)
 
         self.pyArgsString = f"({', '.join(params)})"
+        # __bool__ and __nonzero__ need to be defined as returning int for SIP, but for Python
```
echoix, line 540: `__nonzero__` is python 2, is there still references to that?

echoix:
Note that typealias should come from typing_extensions for Python 3.9 (minimum supported version here), as it was only introduced in 3.10.

It's possible to use type aliases on Python 3.9, just without using the semicolon and TypeAlias, and directly using equal the thing it is aliasing.

etgtools/extractors.py
lojack5: Yeah, the sip code still adds a `__nonzero__` to anything with a `__bool__`.  I didn't remove those since that isn't the focus of this PR.

lojack5:
And yes, `typing_extensions` is already required for < Python 3.10.

echoix:
What started my interest in improving wxPython typing a couple weeks ago was specifically because of the typing for Point and Size. So I'm interested in any improvement for that.

I remember trying hard to find where the conversion from class to tuple happens, but it wasn't clear yet.
This is close to the behaviour I was looking for https://github.com/wxWidgets/Phoenix/blob/c3092be458491c6da89a5a0f0172334e167fcaba/src/wxpy_api.sip#L487

https://github.com/wxWidgets/Phoenix/blob/c3092be458491c6da89a5a0f0172334e167fcaba/etg/gdicmn.py#L243-L252

lojack5:
They're in the sip generating files as far as I can tell:
 - Colour: https://github.com/wxWidgets/Phoenix/blob/c3092be458491c6da89a5a0f0172334e167fcaba/etg/colour.py#L195
 - Position: https://github.com/wxWidgets/Phoenix/blob/c3092be458491c6da89a5a0f0172334e167fcaba/etg/position.py#L37
 - Point: https://github.com/wxWidgets/Phoenix/blob/c3092be458491c6da89a5a0f0172334e167fcaba/etg/gdicmn.py#L104
 - Size: https://github.com/wxWidgets/Phoenix/blob/c3092be458491c6da89a5a0f0172334e167fcaba/etg/gdicmn.py#L181
 - Rect: https://github.com/wxWidgets/Phoenix/blob/c3092be458491c6da89a5a0f0172334e167fcaba/etg/gdicmn.py#L262
 - RealPoint: https://github.com/wxWidgets/Phoenix/blob/c3092be458491c6da89a5a0f0172334e167fcaba/etg/gdicmn.py#L325
 
 So I guess adding things for *all* of these is on the table. I'll see if there's a more general way to do this, instead of hard-coding all of them.

swt2c:
> Maybe.
> 
> Does the conversion added happen _anywhere_ a `wx.DateTime` is accepted as an input to _any_ method, or only `wx.DateTime` method? For example does the conversion happen on `wx.adv.CalendarCtrl.__init__`'s `date` parameter? If so I might be able to do something with that. If not, then it's a lot more special casing that it might be pretty hard to get right.

Yes, it does.  Passing datetime.date to wx.adv.CalendarCtrl works.

lojack5:
Gotcha, I'll start working on a solution that should encompass all types with a `.convertFromPyObject` attribute.  I did a quick glance through and it doesn't seem anything uses the similar-but-reversed `.convertToPyObject`.

lojack5:
Ok, I'm ready for review of this.  In particular if @timrid can find anymore edge cases to fixup, now would be a great time to find them :).  Otherwise, I'm pretty sure this handles most cases correctly now.  There are still some edge cases I found that aren't exactly easy to fix, but from what I can tell they are on rarely used interfaces.

Still would like a yes/no on the `None` for `parent` changes, just a look though to make sure classes that have `parent` marked as `Optional[...]` do in fact accept a `None` for that argument.

lojack5:
Oh, actually there were two points I was unsure about:
`wx.InputStream` and `wx.OutputStream` both have a `.convertFromPyObject` set on them, although the comments in the cpp code there ask if this is only a type-check.  I couldn't find precise documentation on exactly what python type would be deemed acceptable as an automated conversion here.  Is it a specific python type? Or is it a more general "file-like" object - ie: supports a specific interface, like "must have `read`, `tell`, and `seek`?  Possibly something that I could solve with a `typing.Protocol` type-hint there.

timrid:
I downloaded the `wxPython-wheel-win64-py3.13-x64` wheel from https://github.com/wxWidgets/Phoenix/actions/runs/12824937831 and tested it in my codebase. My findings:

1. `wx.Yield()`, `wx.MessageBox(...)` and many more should have no `self` parameter.
image

2. In `wx.Menu()` the Method `DestroyItem` is missing. It was renamed to `Destroy`, but the comment is still correct:
image

2. In `wx.ComboBox()` there is a similar problem with `GetTextSelection` that is renamed to `GetSelection`:
image


4. Missing `GetClientData()` in `wx.Choice()`, `wx.ListBox()`, ... because it is missing in `wx.ItemContainer()`
image

5. I could not see any problem with `parent` set to `None` any more :)

lojack5:
Thanks, I know what's happening with the functions (vs methods) - I can fix that quickly.  The others, not so sure why they're getting renamed, so I'll have to look into that.

lojack5:
Ok, those should be sorted now if you want to take another whirl at it (once the CI jobs complete).

The only one I couldn't reproduce was 3):
type-stub for `ListBox`:
image
and for `ItemContainer`:
image

lojack5:
Also - I take back everything else working as intended - types that are automatically converted are only partially covered, because each generating file is fired up in a separate Python instance, meaning for example the `wx.Point` -> `wx.Point | tuple[int, int]` conversion only happens in the the generating file that defines `wx.Point` itself.  This one might be tricky, because I need the scripts to remember these conversions *between* instances of it running.

lojack5:
Sorry for the delay on this was having some computer issues (frequent BSOD) that I had to fix.  Should get back into finishing this up.

swt2c:
Hey @lojack5 think you'll have time to finish this up soon, or should we merge what you have so far?

lojack5:
Well, I got super busy.  Let me rebase and resolve errors, then you can merge what's here.  The thing not addressed is the type-hints related to automatic type-conversions (ex `tuple[int,int,int]` to `wx.Colour`). For information if anyone want's to try to tackle that part - the issue stems from the fact that each generating file is run in its own python instance. This means that I/we need to implement some sort of cache file for this information, run a quick pass through everything once to collect that information, then do the full actual pass using that cached information.

Metallicow:
@lojack5 look I fixed thousands of your questions in a single ME. Thank you, but some folks wont

PR #2673, opened 2025-01-15
CI: only build the oldest and newest Pythons on PRs

swt2c:
It seems a bit excessive and unnecessary to build for all Pythons on every PR.  Still build for all Pythons on commits.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -113,6 +113,10 @@ jobs:
             architecture: x86
           - os: macos-13
             architecture: x86
+          # Only build oldest and newest Pythons on PRs
+          - python-version: ${{ github.event_name == 'pull_request' && '3.10' }}
+          - python-version: ${{ github.event_name == 'pull_request' && '3.11' }}
+          - python-version: ${{ github.event_name == 'pull_request' && '3.12' }}
 
     env:
       VERSION: ${{ needs.build-source-dist.outputs.VERSION }}


echoix (approved):
Totally makes sense. Would you be open to let one pre-release Python version in some cpu architectures too when it's coming closer to releases? Especially once the abi is stable? I think it could be useful given the nature of the project, that would be used by various other projects.


Talking about prerelease pythons, does this project use multiprocessing, and that has some places that don't work well with the start method that isn't fork? (Like on Windows, where spawn is used). One big change in 3.14 is that the default start method for multiprocessing will not be fork anymore where it was before. On the GRASS GIS project that I maintain, this brings us a lot of work, but will benefit all OSes

swt2c:
> Totally makes sense. Would you be open to let one pre-release Python version in some cpu architectures too when it's coming closer to releases? Especially once the abi is stable? I think it could be useful given the nature of the project, that would be used by various other projects.

Yes, we usually do add the pre-release when it goes ABI stable.

> Talking about prerelease pythons, does this project use multiprocessing, and that has some places that don't work well with the start method that isn't fork? (Like on Windows, where spawn is used). One big change in 3.14 is that the default start method for multiprocessing will not be fork anymore where it was before. On the GRASS GIS project that I maintain, this brings us a lot of work, but will benefit all OSes

I don't think wxPython uses multiprocessing for anything, but thanks for mentioning it as I do use it in some of my other projects.

PR #2680, opened 2025-01-24
Implement __iter__ for wxList iterator classes

swt2c:
This fixes being able to use these classes in for loops, for example. Specifically it fixes the cmdproc tests with Python 3.13.1.

### DIFF
--- a/etgtools/tweaker_tools.py
+++ b/etgtools/tweaker_tools.py
@@ -1055,6 +1055,11 @@ public:
         if (PyErr_Occurred())
             return NULL;
     %End
+
+    PyObject* __iter__();
+    %MethodCode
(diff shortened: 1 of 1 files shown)


DietmarSchwertberger:
@swt2c : Yes, that fixes the issue. Thanks. I will leave the workaround for some time, though.

swt2c:
> I would have expected both to use the same iterator.

I think it's probably the difference between `iter(x)`, which worked, and `iter(iter(x))`, which did not work, where x is a wxList class.

RobinD42:

https://discuss.wxpython.org/t/widget-inspection-tool-broken-in-4-2-2/40184/3

Issue #2422, opened 2023-06-21
Installing wxPython 4.1.2 on OpenSuSE 15.5 ends with missing wx/setup.h file


JacekRuzyczka:
**OpenSuSE 15.5 Linux (x86_64)**:
**4.2.1 via PyPi**:          
**3.11.2-150400.9.5.3 via official OpenSuSE repo**:            

**File wx/setup.h missing when trying to install wxPython by using pip (despite being installed on my machine)**

<details>
Command: `pip wheel -v wxPython-4.2.1.tar.gz  2>&1 | tee build.log`

Error message:
```
  Finished command: build_wx (1m21.690s)
  Running command: build_py
  Checking for /tmp/pip-req-build-2qaqrik2/bin/waf-2.0.24...
  "/usr/bin/python3.11" /tmp/pip-req-build-2qaqrik2/bin/waf-2.0.24 --wx_config=/tmp/pip-req-build-2qaqrik2/build/wxbld/gtk3/wx-config --gtk3 --python="/usr/bin/python3.11" --out=build/waf/3.11/gtk3 configure build
  Setting top to                           : /tmp/pip-req-build-2qaqrik2
  Setting out to                           : /tmp/pip-req-build-2qaqrik2/build/waf/3.11/gtk3
  Checking for 'gcc' (C compiler)          : /usr/bin/gcc
  Checking for 'g++' (C++ compiler)        : /usr/bin/g++
  Checking for program 'python'            : /usr/bin/python3.11
  Checking for python version >= 3.7.0     : 3.11.2
  python-config                            : /usr/bin/python3.11-config
  Asking python-config for pyext '--cflags --libs --ldflags' flags : yes
  Testing pyext configuration                                      : yes
  Finding libs for WX                                              : yes
  Finding libs for WXADV                                           : yes
  Finding libs for WXSTC                                           : yes
  Finding libs for WXHTML                                          : yes
  WARNING: Unable to find setup.h in /tmp/pip-req-build-2qaqrik2/build/wxbld/gtk3/lib64/wx/include/gtk3-unicode-3.2, assuming wxUSE_GLCANVAS is not available.
  Finding libs for WXGL                                            : yes
  WARNING: Unable to find setup.h in /tmp/pip-req-build-2qaqrik2/build/wxbld/gtk3/lib64/wx/include/gtk3-unicode-3.2, assuming wxUSE_WEBVIEW is not available.
  Finding libs for WXWEBVIEW                                       : yes
[...]

Issue #1067, opened 2018-10-31
installing / rebuilding phoenix wxPython fails: wx/setup.h not found


chrismaes87:
**Operating system**: opensuse leap 15.0
**wxPython version & source**:   master as of 29/10/2018: ceb2f467, also tested with versions 4.0.1 and 4.0.3
**Python version & source**:           python2.7 from opensuse repositories

**Description of the problem**:

Since there is no package or wheel available for opensuse 15.0, I tried setting up wxPython in muliple manners:

- following the guide here: https://wxpython.org/blog/2017-08-17-builds-for-linux-with-pip/ ; pip compilation fails on both version 4.0.3 and 4.0.1 (wx/setup.h not found)
- following the readme on the git repository: https://github.com/wxWidgets/Phoenix/blob/master/README.rst (using `python build.py dox etg --nodoc sip build`)

both options fail at the same level:

 ```[ 11/897] Compiling sip/cpp/sip_corewxMDIChildFrameBase.cpp
In file included from ../../../../ext/wxWidgets/include/wx/defs.h:45:0,
                 from ../../../../ext/wxWidgets/include/wx/wx.h:14,
                 from ../../../../wx/include/wxPython/wxpy_api.h:41,
                 from ../../../../sip/cpp/sipAPI_core.h:20434,
                 from ../../../../sip/cpp/sip_corewxMDIChildFrameBase.cpp:10:
../../../../ext/wxWidgets/include/wx/platform.h:148:10: fatal error: wx/setup.h: No such file or directory
 #include "wx/setup.h"
```

When running `python build.py dox etg --nodoc sip build` I traced the following:

- setup.h is generated: **config.status: creating lib/wx/include/gtk3-unicode-3.1/wx/setup.h**
- **build_wx** succeeds (correctly using `-I ...build/wxbld/gtk3/lib/wx/include/gtk3-unicode-3.1`)
- **configure** step seems to complete correctly: 

```
Setting top to                           : /home/chris/workspace/Phoenix 
Setting out to                           : /home/chris/workspace/Phoenix/build/waf/2.7/gtk3 
Checking for 'gcc' (C compiler)          : /usr/bin/gcc 
Checking for 'g++' (C++ compiler)        : /usr/bin/g++ 
Checking for program 'python'            : /usr/bin/python 
Checking for python version >= 2.7.0     : 2.7.14 
python-config                            : /usr/bin/python-config 
[...]

Issue #558, opened 2017-09-28
openSUSE virtualenv build failed


jaredd:
$ uname -a
Linux linux-xdol 4.12.9-1-default #1 SMP PREEMPT Fri Aug 25 07:08:45 UTC 2017 (f2ab6ba) x86_64 x86_64 x86_64 GNU/Linux
$ cat /etc/os-release
cat /etc/os-release 
NAME="openSUSE Tumbleweed"
\# VERSION="20170924"
ID=opensuse
ID_LIKE="suse"
VERSION_ID="20170924"
PRETTY_NAME="openSUSE Tumbleweed"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:tumbleweed:20170924"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"

$ git describe
wxPython-4.0.0b2-13-g88ed1696

Custom build from source

$ python --version
Python 2.7.13

Stock python build (in virtualenv)

$ pwd
/home/jaredd/pyvenvs/gnuradio3.7/src/Phoenix
$ python build.py --prefix=$VIRTUAL_ENV --build_dir=$PWD/build --gtk3 -v dox etg --nodoc
Will build using: "/home/jaredd/pyvenvs/gnuradio3.7/bin/python"
2.7.13 (default, Jan 03 2017, 17:41:54) [GCC]
Python's architecture is 64bit
cfg.VERSION: 4.0.0rc1

Running command: dox
Checking for /home/jaredd/pyvenvs/gnuradio3.7/src/Phoenix/bin/doxygen-1.8.8-linux...
bash ./regen.sh xml
Finished command: dox (17.429s)
Running command: etg
"/home/jaredd/pyvenvs/gnuradio3.7/bin/python" etg/geometry.py --sip --nodoc
Traceback (most recent call last):
  File "etg/geometry.py", line 145, in <module>
    run()
  File "etg/geometry.py", line 54, in run
    c.find('SetPolarCoordinates').ignore()
  File "/home/jaredd/pyvenvs/gnuradio3.7/src/Phoenix/etgtools/extractors.py", line 99, in find
    (head, self.__class__.__name__, self.name))
etgtools.extractors.ExtractorError: Unable to find item named 'SetPolarCoordinates' within ClassDef named 'wxPoint2DDouble'
[...]

PR #2685, opened 2025-02-02
Fix bundled wxWidgets build on OpenSUSE

swt2c:
OpenSUSE has defined its libdir to be 'lib64', which is mismatched with wxWidgets' in-place wx-config, which expects 'lib.'  Work around this by unsetting the CONFIG_SITE envvar (which enables the OpenSUSE customizations) when configuring wxWidgets.

Fixes: https://github.com/wxWidgets/Phoenix/issues/558
Fixes: https://github.com/wxWidgets/Phoenix/issues/1067
Fixes: https://github.com/wxWidgets/Phoenix/issues/2422
Fixes: https://github.com/wxWidgets/Phoenix/issues/2532

### DIFF
--- a/buildtools/build_wxwidgets.py
+++ b/buildtools/build_wxwidgets.py
@@ -370,13 +370,19 @@ def main(wxDir, args):
                 if os.path.exists(frameworkRootDir):
                     shutil.rmtree(frameworkRootDir)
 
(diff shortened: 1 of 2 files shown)


mcepl:
I believe this https://discuss.wxpython.org/t/cant-build-wxpython-4-2-2-on-some-suse-distributions/40176 could be relevant here.

PR #2716, opened 2025-03-14
Fix assert_ to assertTrue

komoto48g:
Fixes #NNNN

This PR fixes the wx.py.tests modules so they run without errors.
Note that this does not mean that the tests pass. There are still some test failures due to recent Python updates.

Fix the test for both cases where `__builtins__` is a dict or a module.
Fix AttributeError: `sys.ps2` which exists only in interactive mode and does not exist when running a script.

### DIFF
--- a/wx/py/tests/test_interpreter.py
+++ b/wx/py/tests/test_interpreter.py
@@ -10,7 +10,7 @@ from wx.py import interpreter
 """
 These unittest methods are preferred:
 -------------------------------------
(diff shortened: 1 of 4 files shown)


swt2c (approved):

echoix (approved):
Indeed, `assert_` was never documented and apparently removed in 3.11.

For the two other types of changes, I assume they are correct, I didn't analyze further.

Thanks for your work!

PR #2747, opened 2025-05-03
Fix up/down option not working on the initial search

### LINKED ISSUE #2744: The up/down option in PyCrust shell's `FindReplaceDialog` doesn't work.
**wxPython version & source**: 4.2.3 (pypi)
**Python version & source**: 3.12

**Description of the problem**: The up/down option in PyCrust shell's `FindReplaceDialog` doesn't work.

<details>
  <summary>To confirm this issue: (click to expand)</summary>

1. Launch PyCrust (or any wx.py.Py*** scripts that inherits from `wx.py.frame`):
   ```bash
   python -m wx.py.PyCrust
   ```
1. Press [Ctrl+F] to open the Find dialog.
1. Select [Up] direction.
1. Enter text such as "wx" and try searching.
You will see that the first match appears below the cursor position.

</details>

komoto48g:
The up/down direction was not applied correctly on the first search because the `backward` parameter was `True`.
This patch introduces a dedicated `_OnFindNext()` method to handle that case explicitly.

Fixes #2744

I prefer not to introduce a new public API unless a better name can be proposed.

### DIFF
--- a/wx/py/frame.py
+++ b/wx/py/frame.py
@@ -543,12 +543,7 @@ class Frame(wx.Frame):
                 "Find & Replace", wx.FR_NOWHOLEWORD|wx.FR_REPLACEDIALOG)
         self.findDlg.Show()
 
-    def OnFindNext(self, event,backward=False):
-        if backward and (self.findData.GetFlags() & wx.FR_DOWN):
-            self.findData.SetFlags( self.findData.GetFlags() ^ wx.FR_DOWN )
-        elif not backward and not (self.findData.GetFlags() & wx.FR_DOWN):
-            self.findData.SetFlags( self.findData.GetFlags() ^ wx.FR_DOWN )
-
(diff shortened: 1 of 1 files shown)


swt2c:
Doesn't this change the public API for `OnFindNext` though?  It seems to remove the `backwards` parameter.  Or is that API not really public?

komoto48g:
Thank you for reviewing.
Yeah, the API did change after all...
The confusing part for me is that `OnFindNext(backward=False)` is called both from the menu command and from the `EVT_FIND` event. So I realized that a better solution would be to check `isinstance(event, wx.FindDialogEvent)`.
I'll amend the code, check it, and force-push later ... (sorry for the force-push).

Functions that start with _ aren't documented by Sphinx -- I intended that to mean "not public".

komoto48g:
I force-pushed the changes. So the API doesn't need to be changed after all.

PR #2448, opened 2023-08-21
Float canvas updates

ChrisBarker-NOAA:
Fixed few bugs for newer wxPython, numpy and Python.

Tested with:

- Python 3.10
- numpy 1.25
- wxMac OS-X 11.7.9

Fixes #2320, #2331 

Also got many of the samples working, and cleaned up a bunch of formatting.

I only touched FloatCanvas code.

PR #2749, opened 2025-05-06
Fix floatcanvas Text.ShiftFunDict: convert float to int

### LINKED ISSUE #2748: FLoatCanvas Bitmap & Text always get drawn at (0,0) under Python 3.11
**wxPython version & source**: PyPi wxPython 4.2.3
**Python version & source**:    Python 3.11

Some objects are drawn in FloatCanvas at (0,0) instead of their actual location.

<details>
Here's the root cause:
With any object that's shifted in FloatCanvas (e.g. FloatCanvas.Bitmap) the shift function (ShiftFun) causes conversion of the coordinates from int to float.
This in turn causes the drawing function (part of MemoryDC) to fail, causing all object to be drawn at (0,0)

To see this launch the wxDemo - FloatCanvas
This is the result under Python 3.11:

Image

This is what it looks like under Python 3.9:

Image

</details>

komoto48g:
Fixes #2748

This PR modifies "wx/lib/floatcanvas/FCObjects.py" and is part of #2448, which currently has conflicts.

### DIFF
--- a/wx/lib/floatcanvas/FCObjects.py
+++ b/wx/lib/floatcanvas/FCObjects.py
@@ -11,7 +11,7 @@
 # Tags:         phoenix-port, unittest, documented, py3-port
 #----------------------------------------------------------------------------
 """
-This is where FloatCanvas defines its drawings objects.
+This is where FloatCanvas defines its drawing objects.
 """
 
 import sys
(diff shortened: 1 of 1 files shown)


komoto48g:
Sorry, I hadn't checked the bitmap drawing test. I've confirmed that the bitmap position is also incorrect.
But I just pushed one more commit—could you revert or reopen this? Sorry for the inconvenience.

swt2c:
Can't - you'll have to open a new PR.

komoto48g:
Ok, I will!

urishab:
I’m not sure this fix is gonna cut it. 
The // operator no longer converts to int. 
it just gives the floor of the number but can return float. 
so I think you’re going to have to explicitly convert to int to satisfy all cases.

komoto48g:
I can see that, but it doesn't cover all cases. I also tested the Allow and AllowLine cases and confirmed that they have issues as well. The issue comes from passing a float value to dc.DrawXXX.
I will check some more and submit a PR later and would appreciate it if you could review it.

PR #2818, opened 2025-11-09
Update waf to v2.1.7

swt2c:
This is needed primarily for building for free-threaded Python.

Fixes #NNNN

### DIFF
--- a/build.py
+++ b/build.py
@@ -85,8 +85,8 @@ wxICON = 'packaging/docset/mondrian.png'
 
 # Some tools will be downloaded for the builds. These are the versions and
 # MD5s of the tool binaries currently in use.
-wafCurrentVersion = '2.1.5'
-wafMD5 = '2e7b2166c030dbac3e21891048df10aa'
+wafCurrentVersion = '2.1.8'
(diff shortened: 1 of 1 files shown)


swt2c:
Looks like I introduced a regression in WAF, fix submitted:
https://gitlab.com/ita1024/waf/-/merge_requests/2396

kdschlosser:
I believe there is a new version of WAF that was released a week ago., It should have the fix in it for the issue you were having here.  I hope it does.. It would be nice to use Python without the GIL as I have some database accessing that feeds into the UI that would get a pretty good performance boost from being able to execute code in parallel.. I looked at using subprocesses but because of the need to pickle objects in order to pass them between processes that pretty much wipes out an performance gains gotten from running multiple process. If I was doing a single massive query of the database it would be faster but because I am dealing with lots of small operations there is simply too much overhead to make it worth while. It would be beneficial to run python without a GIL using threading in my use case and wxPython not being able to run on it is the only thing keeping me from using it. I can update the code to use the latest version of WAF but it would be nice to have an official release where it can be installed using Pypi.

PR #2822, opened 2025-11-13
Fix wx.svg to work with cython 3.1 generated code.

### LINKED ISSUE #2821: wx.svg does not work with cython 3.1 generated code
wx.svg fails to work starting from wxpython 4.2.4. Specifically, any call to `SVGimage.RenderToGC` would fail.

The cause of this issue is the switch from cython 3.0 to 3.1. In cython 3.0 enums values were defined at the module level. But starting in cython 3.1 they are only defined in the enum namespace.

Therefore, `wx.svg._nanosvg.SVG_FLAGS_VISIBLE` was accessible in wxpython 4.2.3, but now we need to use `wx.svg._nanosvg.SVGflags.SVG_FLAGS_VISIBLE`.

udifuchs:
Fixes #2821

### DIFF
--- a/unittests/test_dcsvg.py
+++ b/unittests/test_dcsvg.py
@@ -1,7 +1,7 @@
 import unittest
 from unittests import wtc
 import wx
-import sys
+import wx.svg
 import os
 
 fileName = 'svgtest.svg'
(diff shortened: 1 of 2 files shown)


swt2c:
> Isn’t cython used as a build dependency only? Aren’t the build dependencies already isolated between projects?
> 
> do we really need compatibility? Accessing private members is already not intended to be done

The enums are also in the `wx.svg` namespace.

udifuchs:
I just added the missing enums.

Technically, the first commit is no longer necessary, but I think it is more elegant to access enums values by their namespace.

PR #2837, opened 2025-12-16
Bugfix in python-only hypertreelist. Add missing attribute _x_maincol to constructor

### LINKED ISSUE #2828: TreeListMainWindow: AttributeError for _x_maincol when calling GetItemWidth() before CalculatePositions()
**Description**

After updating to **wxPython 4.2.4**, code that previously worked now fails with:

```
AttributeError: 'TreeListMainWindow' object has no attribute '_x_maincol'
```

The error occurs in `TreeListMainWindow.GetItemWidth()`, which accesses `self._x_maincol`. However, this attribute is not initialized in the constructor; it is only set later in `CalculatePositions()`.

If `GetItemWidth()` is called before `CalculatePositions()`, the attribute does not yet exist, causing the exception.

Manually calling `CalculatePositions()` before invoking `GetItemWidth()` avoids the error, but it appears that `_x_maincol` should be initialized during construction to prevent this failure.

**Steps to Reproduce**

1. Create a `TreeListCtrl` and access item width (directly or indirectly) early in the widget’s lifecycle.
2. Call a method that triggers `GetItemWidth()` before `CalculatePositions()` has run.
3. Observe the `AttributeError`.

**Environment**

* wxPython version: 4.2.4
* OS: Windows 11
* Python version: 3.13

**Possible Fix**

Initialize `self._x_maincol` (e.g., to `0`) in the `TreeListMainWindow` constructor.

jmoraleda:
Fixes #2828

### DIFF
--- a/wx/lib/agw/hypertreelist.py
+++ b/wx/lib/agw/hypertreelist.py
@@ -2341,6 +2341,7 @@ class TreeListMainWindow(CustomTreeCtrl):
         self._curColumn = -1    # no current column
         self._owner = parent
         self._main_column = 0
+        self._x_maincol = 0
         self._dragItem = None
 
         self._imgWidth = self._imgWidth2 = 0


echoix:
Is the submodule update expected?

jmoraleda:
Apologies... Again... Hopefully it is fixed now.

PR #2853, opened 2026-02-06
Update sip to v6.15.1

swt2c:
Fixes #NNNN


echoix:
Wouldn't you want dependabot or renovate to open such PRs for you? I prefer renovate, as you can really make it do anything you want even if it's custom only for your project.

PR #2880, opened 2026-04-15
Added a check to not remove control panes that don't have a window yet

### LINKED ISSUE #2865: new LoadPerspective breaks old saved perspective
This error was introduced in wxPython 4.2.3

This is what should look like, but on the newest versions it doesn't appear on screen at all:

Image

After some debugging I found that the difference was in the new way the function LoadPerspective inside the class AuiManger works, in particular this section:

```python

        inexistentPaneIndexes = []
        for paneIndex, p in enumerate(self._panes):
            if p.window:
                if p.IsMinimized():
                    self.MinimizePane(p, False)
                elif p.minimize_mode & AUI_MINIMIZE_POS_MASK == AUI_MINIMIZE_POS_TOOLBAR:
                    # If the pane specifies a minimize toolbar, which would be preserved by the loading
                    # operation rather than using an automatic toolbar which would not be preserved) then 
                    # if pane was minimized before we loaded perspective but the new perspective unminimized
                    # it then its minimize-from-restore icon is still in the toolbar and must be removed
                    minimizeToolbar = self.GetPane(p.minimize_target).window
                    item = minimizeToolbar.FindToolByUserData((ID_RESTORE_FRAME, p.window))
                    if item: # Delete the icon if it was there
                        minimizeToolbar.DeleteTool(item)
            else:
                # The perspective is referencing a non-existing pane that was added nonetheless. This will
                # happen with a notebook that ends up containing no pages
                inexistentPaneIndexes.append(paneIndex)
                
        for paneIndex in reversed(inexistentPaneIndexes): # reverse is required when deleting elements by index
            del self._panes[index]

```

the element "__notebook_0" isn't associated to a window so it matches the condition for inexistent panes and gets deleted

this is the line managed differently

`name=__notebook_0;caption=Transcript;state=2044;dir=3;layer=2;row=0;pos=0;prop=100000;bestw=118;besth=49;minw=-1;minh=-1;maxw=-1;maxh=-1;floatx=902;floaty=968;floatw=118;floath=49;notebookid=0;transparent=255|`

here's the full perspective just for info:

It appears that our perspective doesn't load correctly on any newer version this is the perspective string

`layout2|name=nbRight;caption=;state=768;dir=5;layer=0;row=0;pos=0;prop=100000;bestw=0;besth=0;minw=-1;minh=-1;maxw=-1;maxh=-1;floatx=-1;floaty=-1;floatw=-1;floath=-1;notebookid=-1;transparent=255|name=txtOutput;caption=Transcript;state=264188;dir=6;layer=0;row=0;pos=0;prop=100000;bestw=118;besth=49;minw=-1;minh=-1;maxw=-1;maxh=-1;floatx=902;floaty=968;floatw=118;floath=49;notebookid=0;transparent=255|name=__notebook_0;caption=Transcript;state=2044;dir=3;layer=2;row=0;pos=0;prop=100000;bestw=118;besth=49;minw=-1;minh=-1;maxw=-1;maxh=-1;floatx=902;floaty=968;floatw=118;floath=49;notebookid=0;transparent=255|name=locator_window;caption=Locator;state=264188;dir=6;layer=0;row=0;pos=1;prop=100000;bestw=1759;besth=169;minw=-1;minh=-1;maxw=-1;maxh=-1;floatx=800;floaty=969;floatw=1759;floath=169;notebookid=0;transparent=255|name=Resources;caption=Resources;state=264188;dir=6;layer=0;row=0;pos=2;prop=100000;bestw=821;besth=238;minw=-1;minh=-1;maxw=-1;maxh=-1;floatx=847;floaty=941;floatw=821;floath=238;notebookid=0;transparent=255|name=prjTree;caption=Project Tree;state=264188;dir=4;layer=3;row=0;pos=0;prop=100000;bestw=100;besth=99;minw=-1;minh=-1;maxw=-1;maxh=-1;floatx=-16;floaty=557;floatw=100;floath=99;notebookid=-1;transparent=255|dock_size(5,0,0)=10|dock_size(3,2,0)=214|dock_size(6,0,0)=576|dock_size(4,3,0)=262|"`

I don't really know how "__notebook_0" is generated in the first place and it seems it's necessary for the perspective to load correctly

AlessandroMattiazzi:
This issue broke retrocompatibility with old saved perspectives, this PR fixes it

Fixes #2865

### DIFF
--- a/wx/lib/agw/aui/framemanager.py
+++ b/wx/lib/agw/aui/framemanager.py
@@ -5369,8 +5369,11 @@ class AuiManager(wx.EvtHandler):
                         minimizeToolbar.DeleteTool(item)
             else:
                 # The perspective is referencing a non-existing pane that was added nonetheless. This will
-                # happen with a notebook that ends up containing no pages
-                inexistentPaneIndexes.append(paneIndex)
+                # happen with a notebook that ends up containing no pages.
+                # Notebook control panes (for example "__notebook_0") are expected to not have windows 
+                # at this point because they will be created later by Update(), so we don't delete them.
+                if not p.IsNotebookControl():
+                    inexistentPaneIndexes.append(paneIndex)
                 
         for paneIndex in reversed(inexistentPaneIndexes): # reverse is required when deleting elements by index
             del self._panes[index]


swt2c:
@jmoraleda any comments here?

jmoraleda:
@swt2c, thank you for asking. I am traveling and I won't be able to look at this properly until next week.

jmoraleda:
Hello @swt2c and @AlessandroMattiazzi 

I have reviewed the issue #2865 and confirm that the issue reported is a reproducible bug. Furthermore I confirm this PR fixes it. But I recommend that we don't merge this PR, and we merge #2882 instead, because in the process of investigating the issue, I realized that the problem was a side effect of a larger issue and we can address both problems simultaneously while reducing the code base. Please see my detailed rationale there.

Thank you both of you.

AlessandroMattiazzi:
Thank you for the support, analysis and PR review I'll close this PR in favor of #2882

PR #2895, opened 2026-05-16
Shell improvements

DietmarSchwertberger:
This PR should improve wx.py.shell:
 - auto-completion on pressing tab
 - auto-completion on e.g. `import `
 - auto-completion on e.g. `sys.path.a`
 - automatic calltip on e.g. completion of `sys.path.a` to `sys.path.append(`
 - calltip also on class instantiation, not only on function calls
 - context menu entry "Copy History"

@newville , @komoto48g : would you mind testing?

I started this when I saw the new, improved shell features of the recent Python versions. I was hoping there was a bit more abstraction...

P.S.: I'm testing / developing this on Python 3.12 and 3.14. I think that I'm not using recent features, but need to test.

### DIFF
--- a/wx/py/introspect.py
+++ b/wx/py/introspect.py
@@ -174,35 +174,36 @@ def getCallTip(command='', locals=None):
         pass
     tip1 = ''
     argspec = ''
+    doc = ''
     obj = inspect.unwrap(obj)
     if inspect.isbuiltin(obj):
         # Builtin functions don't have an argspec that we can get.
         pass
-    elif inspect.isfunction(obj):
+    elif callable(obj):
         # tip1 is a string like: "getCallTip(command='', locals=None)"
         try:
-            argspec = str(inspect.signature(obj)) # PY35 or later
-        except AttributeError:
-            argspec = inspect.getfullargspec(obj)
-            argspec = inspect.formatargspec(*argspec)
-        if dropSelf:
-            # The first parameter to a method is a reference to an
-            # instance, usually coded as "self", and is usually passed
-            # automatically by Python; therefore we want to drop it.
-            temp = argspec.split(',')
-            if len(temp) == 1:  # No other arguments.
-                argspec = '()'
-            elif temp[0][:2] == '(*': # first param is like *args, not self
-                pass
-            else:  # Drop the first argument.
-                argspec = '(' + ','.join(temp[1:]).lstrip()
-        tip1 = name + argspec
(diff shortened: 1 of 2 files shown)


RobinD42:

https://discuss.wxpython.org/t/users-of-shell/40437/3

newville:
@DietmarSchwertberger Thanks for working on this.

On macOS with Python 3.13.13, I get

```
>>> import wx.py as py
...
 File "..../lib/python3.13/site-packages/wx/py/shell.py", line 28, in <module>
    from _pyrepl import _module_completer
ImportError: cannot import name '_module_completer' from '_pyrepl' (/.../lib/python3.13/_pyrepl/__init__.py)
```

Sorry, I haven't been able to investigate further.

DietmarSchwertberger:
Thanks. It seems _pyrepl is available on 3.13, but not _module_completer.
I have modified to check for ModuleNotFoundError and also ImportError.
The full autocompletion including imports is therefore only available on 3.14.

DietmarSchwertberger:
To be done:
 - handle cases like `l[1].`
 - handle `.`

update: they are ignored now

komoto48g:
Thank you for working on this PR @DietmarSchwertberger.
I would like to leave some comments.

1. auto-completion on pressing tab
   Using the Tab key to trigger completion is natural and consistent with the standard PyREPL behavior.
2. auto-completion on e.g. `import`
   Nice. It would also be nice if `from x import y` statements were supported.
3. auto-completion on e.g. `sys.path.a`
   When a  `.`  and the `a` key is pressed, auto-completion is invoked. Do you have more ideas for improving this behavior?
4. automatic calltip on e.g. completion of `sys.path.a` to `sys.path.append(`
   I think using an icon instead of ( to indicate callables is more natural and visually intuitive.
   The `stc.StyledTextCtrl` completion function has functionality to set icons for each completion item.
5. calltip also on class instantiation, not only on function calls
6. context menu entry "Copy History"
   These seem good. Maybe a separate PR would be better.

~~I tested it with Python 3.13.14 and wxPython 4.3.1 msw (Phoenix) with wxWidgets 3.3.3. However, module completion does not work well.
I guess `_pyrepl` has changed since this PR was created, and `_pyrepl._module_completer` has been removed. So, relying on a private module may cause compatibility issues in the future.~~ Never mind. You already mentioned it above. I’ll install Python 3.14 and test it later.

PR #2897, opened 2026-05-19
Set rpath in link arguments rather than LD_RUN_PATH

swt2c:
This should fix wxPython wheel installs when using pyenv as pyenv sets an rpath link argument, which was overriding LD_RUN_PATH.  Instead, set rpath to $ORIGIN in a link argument, which should be additive with the pyenv one(s).


RobinD42:

https://discuss.wxpython.org/t/installing-on-linux-under-pyenv/40429/11

PR #2903, opened 2026-06-04
Use ccache to speed up CI builds

swt2c:


echoix (changes requested):
You’d want to avoid using that cache for releases, tags. And maybe not store it (write) from PRs, in order to avoid cache poisoning attacks. If you run (manually) the zizmor linter, you’ll get an idea of what is problematic, and see the autofix it might suggest.

swt2c:
> You’d want to avoid using that cache for releases, tags. And maybe not store it (write) from PRs, in order to avoid cache poisoning attacks. If you run (manually) the zizmor linter, you’ll get an idea of what is problematic, and see the autofix it might suggest.

Yeah, I'm aware of those issues.  This PR is far from complete, so will mark as draft.

swt2c:
@echoix what are your thoughts on whether the cache is used for snapshot builds?

echoix:
> @echoix what are your thoughts on whether the cache is used for snapshot builds?

Since they are now published, and are like « rc », I would consider them like releases, like a practice on how to release. If they are copied to the extras on the web server, I’d be extra careful (pun intended) so that we are not publicly hosting a potentially poisoned artifact.

I think the « snapshot » builds should be using the same process as releases, so we are getting practice on doing it without intervention on a personal (and thus untrusted) computer. At least on CI, it is possible to look at what happened with multiple eyes

echoix:
I have no problem on populating a cache from the default branch, and speeding up PR builds though. And I would probably impose a hard expiration even if still actively used on all these caches to limit how much time a « bad » cache could affect. Like a week or two max.

swt2c:
OK, we won't use cache for snapshots.

swt2c:
@echoix I think this is ready.  Any thoughts on the final implementation?  It should save the cache only during non-PRs but only restore it during PRs.

.github/workflows/ci-build.yml
```diff
@@ -224,6 +224,25 @@ jobs:
...
+      - name: Install ccache
+        if: ${{ !startsWith(github.ref, 'refs/tags/') && runner.os != 'Windows' }}
+        uses: hendrikmuhs/ccache-action@v1.2.23
+        with:
+          key: ${{ github.job }}-${{ matrix.os }}-${{ matrix.python-version }}-${{ matrix.architecture }}
+          restore: ${{ github.event_name == 'pull_request' }}
+          save: ${{ github.event_name != 'pull_request' }}
+
+      - name: Setup ccache
+        if: ${{ !startsWith(github.ref, 'refs/tags/') }}
```
echoix: This means it runs on windows, but is it installed another way than the action above? I know it shouldn't match a branch of the if though

echoix (approved):
What's remaining as an interrogation for me is if the cache keys are only a prefix/suffix, and other info are added that would make subsequent saves be uploaded and used (instead of only keeping the first and oldest one alive and alive). I don't see another way of having them rotated/expired (except from usual pruning when cache entries are unused).

I opened one job run, a macOS one, and the post ccache step listing the statistics showed that it used 0.4/0.5 GB limit set. Assuming this is already a compressed size, and nothing else saves to cache, that represents 25 cache entries for in the 10 GB.


With what I understand, the snapshots tags don't run this, so it won't duplicate the number of saved entries (and it wouldn't fall back to be used anyways, since it wouldn't come from the default branch).

Looking at what is in the cache, I see that some entries (dependencies) are over two months old, and that cache was used less than 24 h ago. I didn't look specifically at what action uses that, but I know that for another project using micromamba, if there's a cache hit, the environment would not be resolved again, thus not checking if there's something to update in the transitive dependencies.  Either clear all cache items manually with the gh cli after some time, or add a time part to the cache key that would invalidate after a week or a month.

But my conclusion is: start with this, and take a look on how it behaves. If the amount of cache entries take too much space, you'll end up always uploading new entries without having the chance to use them once. See cache thrashing.

.github/workflows/ci-build.yml
swt2c: Good point, we can skip the step completely on runner.os = Windows to make it clearer.

opq0pqo (approved):

swt2c:
Unfortunately, it seems ccache is not helping very much.  Not sure why.  Only 3-6% cache hits, looking at:
https://github.com/wxWidgets/Phoenix/actions/runs/31186768304

echoix:
> Unfortunately, it seems ccache is not helping very much.  Not sure why.  Only 3-6% cache hits, looking at:
> 
> https://github.com/wxWidgets/Phoenix/actions/runs/31186768304

I have a strong suspicion that, if it is pip that is building the wheel from the source dist, then it will use a unique path to build the wheel into.




I have already seen similar things with R packages in CI, but there I had to loosen the time precision and folder hashes mid run, after the compilation of the project. 

https://github.com/echoix/grass-addons/blob/393a0885773182a87ff03d3965af2e4203de3e3b/.github/workflows/ci.yml#L128-L133


It is also possible to set the verbosity higher, so there is more info in the stats. Set it in the ccache action like I did in the same file linked

PR #2906, opened 2026-06-14
CI: Add Ubuntu 26.04 to matrix

echoix:
I wanted to set up a new dev environment, and found out that there were no prebuilt wheels for Ubuntu 26.04. This is an attempt to see what else breaks when naively adding ubuntu 26.04 runner images to the os matrix. These runner images are in preview, so they are available.

### DIFF
--- a/.github/workflows/build-linux-wheels.yml
+++ b/.github/workflows/build-linux-wheels.yml
@@ -57,7 +57,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04, ubuntu-24.04 ]
+        os: [ ubuntu-22.04, ubuntu-24.04, ubuntu-26.04 ]
         python-version: [ '3.10', '3.11', '3.12', '3.13', '3.14' ]
         architecture: [ 'x64' ]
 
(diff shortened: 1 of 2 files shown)


echoix:
@swt2c Could you take a look at this, and possibly either do a release or upload newer binaries?

We would want to update a container to Ubuntu 26.04, but compiling wxPython each time is boring. We are waiting for these wheels to be available

swt2c:
Would you want another 4.2.x release?  I'm not sure we're ready for a 4.3.x release yet.

echoix:
Sure, any patch release is fine

swt2c:
@echoix I uploaded v4.2.5 wheels to the GitHub Release: https://github.com/wxWidgets/Phoenix/releases/tag/wxPython-4.2.5

This is the plan going forward - trying to move content off Robin's server.

echoix:
Hmm, ok, I'll see how we could change our build script to not use the server as a source for pip install. I'm not sure how it'll handle it, but we could at least try

swt2c:
> Hmm, ok, I'll see how we could change our build script to not use the server as a source for pip install. I'm not sure how it'll handle it, but we could at least try

`pip install https://github.com/wxWidgets/Phoenix/releases/download/wxPython-4.2.5/wxpython-4.2.5+ubuntu2604-cp314-cp314-linux_x86_64.whl` doesn't work?

PR #2918, opened 2026-07-23
Fix wxAuiDefaultTabArt instantiation

itmo153277:
wxWidgets docs are missing implementation details for wxAuiFlatTabArt. Because of this, wxAuiFlatTabArt is mislabelled as abstract


swt2c:
On this one, I wonder if we should discuss with wxWidgets whether the interface files should be updated/corrected?

PR #2921, opened 2026-07-24
Add AuiNativeTabArt

itmo153277:
Based on:
https://github.com/wxWidgets/wxWidgets/blob/v3.3.3/include/wx/aui/tabart.h#L461-L475
https://github.com/wxWidgets/wxWidgets/blob/v3.3.3/interface/wx/aui/auibook.h#L1274-L1293

### DIFF
--- a/etg/auibook.py
+++ b/etg/auibook.py
@@ -15,6 +15,8 @@ MODULE    = "_aui"
 NAME      = "auibook"   # Base name of the file to generate to for this script
 DOCSTRING = ""
 
+INCLUDES = [ 'tabartnative' ]
+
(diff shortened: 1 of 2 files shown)


itmo153277:
Maybe just replacing the typedef with a classdef would be better? It should still pick up the correct implementation, and would be cleaner

swt2c:
> Maybe just replacing the typedef with a classdef would be better? It should still pick up the correct implementation, and would be cleaner

Yes, I was thinking along those lines.

itmo153277:
@swt2c I created a new PR (#2923) that uses that method. You can choose whichever you prefer

swt2c:
Merged #2923 so this one is not needed, thanks.

PR #2922, opened 2026-07-25
Add missing transfer attributes for SetArtProvider

itmo153277:
This avoids triggering segmentation faults when setting custom art providers

### DIFF
--- a/etg/auibar.py
+++ b/etg/auibar.py
@@ -38,6 +38,7 @@ def run():
     c = module.find('wxAuiToolBar')
     assert isinstance(c, etgtools.ClassDef)
     tools.fixWindowClass(c)
+    c.find('SetArtProvider.art').transfer = True
 
 
     c = module.find('wxAuiToolBarEvent')
--- a/etg/auibook.py
+++ b/etg/auibook.py
@@ -57,6 +57,7 @@ def run():
     c = module.find('wxAuiTabContainer')
     tools.ignoreConstOverloads(c)
     c.find('GetPages').ignore()
+    c.find('SetArtProvider.art').transfer = True
 
     notebook_page_equality_code = """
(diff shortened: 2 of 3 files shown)


etg/ribbon_control.py
```diff
@@ -38,6 +38,7 @@ def run():
     c.find('DoGetNextSmallerSize').ignore(False)
     c.find('DoGetNextLargerSize').ignore(False)
     tools.fixWindowClass(c)
+    c.find('SetArtProvider.art').transfer = True
```
swt2c: I don't think this one should be transfer.  It doesn't look like `wxRibbonControl` deletes the m_art (perhaps it should), so this would result in a memory leak.

etg/ribbon_page.py
```diff
@@ -36,6 +36,7 @@ def run():
     c = module.find('wxRibbonPage')
     assert isinstance(c, etgtools.ClassDef)
     tools.fixWindowClass(c)
+    c.find('SetArtProvider.art').transfer = True
```
swt2c: Same here, I think.

etg/ribbon_panel.py
```diff
@@ -39,6 +39,7 @@ def run():
     assert isinstance(c, etgtools.ClassDef)
     tools.fixWindowClass(c)
     tools.ignoreConstOverloads(c)
+    c.find('SetArtProvider.art').transfer = True
```
swt2c: Same here.

etg/ribbon_control.py
itmo153277: My bad. RibbonBar owns the art provider and just propagates to everyone else. I just assumed that it would clone it, but it just passes the pointer. Currently RibbonBar already has the attribute, so the ribbon part should be good as is.

PR #2925, opened 2026-07-27
wx.lib.inspection dark mode issues [Issue #2917]

### LINKED ISSUE #2917: wx.lib.inspection dark mode issues.
Operating system = Windows 11.
wxPython version & source =  I’m using a fork of [wxPython](https://pypi.org/project/wxPythonPyRx/)  based off Commit 2cddce0, no changes 
Python 3.14

Some controls are not dark, it should be determined if it's the control so it can be reported to wxWidgets 
 
Image

<details>

```python
import wx
import wx.lib.inspection

class BugReproductionApp(wx.App):
    def OnInit(self):
        self.MSWEnableDarkMode(1) 
        self.frame = wx.Frame(None, title="Main App Frame", size=(400, 300))
        self.frame.Show()
        wx.lib.inspection.InspectionTool().Show()
        return True
if __name__ == "__main__":
    app = BugReproductionApp()
    app.MainLoop()
```
</details>

CEXT-Dan:
this commit partially fixes the dark mode issues in wx.lib.inspection 
by making the editor window readable when in dark mode

Fixes #2917

### DIFF
--- a/wx/py/editwindow.py
+++ b/wx/py/editwindow.py
@@ -14,10 +14,6 @@ from . import dispatcher
 from .version import VERSION
 
 
-import wx
-
-FACES = None
-
 def get_faces():
     if 'wxMSW' in wx.PlatformInfo:
         # Safely evaluate IsDark() inside a try-except fallback block
--- a/wx/py/editwindow.py
+++ b/wx/py/editwindow.py
@@ -15,13 +15,14 @@ from .version import VERSION
 
 
 def get_faces():
+    
+    try:
+        is_dark = wx.SystemSettings.GetAppearance().IsDark()
+    except wx.PyNoAppError:
+        is_dark = False
+        
     if 'wxMSW' in wx.PlatformInfo:
         # Safely evaluate IsDark() inside a try-except fallback block
-        try:
-            is_dark = wx.SystemSettings.GetAppearance().IsDark()
-        except wx.PyNoAppError:
-            is_dark = False
-
         return { 
             'times'     : 'Times New Roman',
             'mono'      : 'Courier New',
--- a/wx/py/editwindow.py
+++ b/wx/py/editwindow.py
@@ -14,7 +14,11 @@ from . import dispatcher
 from .version import VERSION
 
 
+import wx
+
 if 'wxMSW' in wx.PlatformInfo:
+    is_dark = wx.SystemSettings.GetAppearance().IsDark()
+
     FACES = { 'times'     : 'Times New Roman',
               'mono'      : 'Courier New',
               'helv'      : 'Arial',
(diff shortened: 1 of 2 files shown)


CEXT-Dan:
sorry, actually, this does not work with the sample I posted in #2917

CEXT-Dan:
now the sample works

wx/py/editwindow.py
```diff
@@ -14,55 +14,70 @@
...
-              'mono'      : 'Courier',
-              'helv'      : 'Helvetica',
-              'other'     : 'new century schoolbook',
-              'size'      : 12,
-              'lnsize'    : 10,
-              'backcol'   : '#FFFFFF',
-              'calltipbg' : '#FFFFB8',
-              'calltipfg' : '#404040',
-            }
+import wx
```
swt2c: wx is already imported on line 5.

wx/py/editwindow.py
```diff
@@ -72,6 +87,8 @@ class EditWindow(stc.StyledTextCtrl):
     def __init__(self, parent, id=-1, pos=wx.DefaultPosition,
                  size=wx.DefaultSize, style=wx.CLIP_CHILDREN | wx.SUNKEN_BORDER):
         """Create EditWindow instance."""
+        global FACES
```
swt2c: Instead of using a `global`, why not just use `self.FACES = get_faces()`?

wx/py/editwindow.py
```diff
@@ -14,55 +14,70 @@
...
+
+FACES = None
+
+def get_faces():
+    if 'wxMSW' in wx.PlatformInfo:
+        # Safely evaluate IsDark() inside a try-except fallback block
+        try:
+            is_dark = wx.SystemSettings.GetAppearance().IsDark()
+        except wx.PyNoAppError:
+            is_dark = False
```
swt2c: Can we fix this for platforms other than MSW as well?  Maybe factor out the color setting to be for all platforms, since the colors are the same for all platforms?

swt2c (changes requested):

wx/py/editwindow.py
CEXT-Dan: Okay, I will make the adjustments. I did try using self.FACES at first, but it did not work. I will try again

wx/py/editwindow.py
swt2c: It looks like you tried to address this, but it seems that you need to move the `is_dark` definition outside of `if 'wxMSW' in wx.PlatformInfo`

wx/py/editwindow.py
```diff
@@ -151,7 +162,7 @@ def setStyles(self, faces):
 
         # Built in styles
         self.StyleSetSpec(stc.STC_STYLE_LINENUMBER,
-                          "back:#C0C0C0,face:%(mono)s,size:%(lnsize)d" % FACES)
+                          "back:#C0C0C0,face:%(mono)s,size:%(lnsize)d" % faces)
```
swt2c, line 166: Shouldn't this be `self.faces`?

wx/py/editwindow.py
CEXT-Dan: oof, I'll fix that

wx/py/editwindow.py
CEXT-Dan: the way I read it is that it was initially a bug, method setStyles(self, faces): has a clear intent, to let the user set the setStyles. Why would the original mix local an global instances of faces?

swt2c: Ooops, I didn't see the `faces` local variable.

swt2c:
Now that we have the unittests up and running in CI/CD, can you either rebase or merge master into this branch so they run?  Also, can you add a unittests for this functionality, if one doesn't already exist?

CEXT-Dan:
Sorry boss, I’m way out of my league here, I changed all the CI/CD on my fork. It’s way out of date. happy to create a test in another PR. I probably need to redo my fork

swt2c:
I can take care of it, if you'd like?

CEXT-Dan:
That would be great.  I made a post on the forum asking for advice on how to contribute, maybe that will help me out in the future

PR #2929, opened 2026-07-30
Fix stale event object in GridWithLabelRenderersMixin paint handlers

jmoraleda:
The three paint handlers (_onPaintRowLabels, _onPaintColLabels, _onPaintCornerLabel) resolved their target window via evt.GetEventObject(), even though each handler is bound directly to one specific label window in __init__ and so already knows which window it is.

Resolving via GetEventObject() requires SIP to look up the event's wxObject* in its C++-to-Python object map. If some unrelated, non- window wxObject is deleted without SIP being notified (e.g. a wx.MenuItem freed by ~wxMenuBase, which does not go through SIP's teardown path) and a new C++ object is later allocated at that same freed address, the map lookup can return a stale Python wrapper of the wrong type for the live paint event. This surfaces as "TypeError: PaintDC(): argument 1 has unexpected type '<OtherClass>'" raised repeatedly on every paint of the affected label window, since the window itself is alive and genuinely painting.

Reported before at: https://discuss.wxpython.org/t/typeerror-when-using-gridwithlabelrenderersmixin/35137

Fix: cache the three label windows in __init__ (the same windows the handlers are bound to) and use the cached references instead of evt.GetEventObject().

### DIFF
--- a/wx/lib/mixins/gridlabelrenderer.py
+++ b/wx/lib/mixins/gridlabelrenderer.py
@@ -26,9 +26,25 @@ class GridWithLabelRenderersMixin(object):
     how the cell renderers work in the main Grid class.
     """
     def __init__(self):
-        self.GetGridRowLabelWindow().Bind(wx.EVT_PAINT, self._onPaintRowLabels)
-        self.GetGridColLabelWindow().Bind(wx.EVT_PAINT, self._onPaintColLabels)
-        self.GetGridCornerLabelWindow().Bind(wx.EVT_PAINT, self._onPaintCornerLabel)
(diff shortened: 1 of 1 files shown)


jmoraleda:
Filed the underlying stale-wrapper bug that this mixin fix defends against: #2931. Confirmed the same mechanism also affects `wx.SizerItem` (matching the `SizerItem` symptom in the originally-linked forum thread) and `wx.ToolBarToolBase`, with repros for all three. This PR stands on its own regardless of how #2931 gets resolved -- it stops trusting `evt.GetEventObject()`, which is unsafe independent of the root cause.

PR #2943, opened 2026-08-13
Configuration Support for building universal2 binaries on newer macOS versions

jaknel:
"build.py --mac_arch=arm64,x86_64 build" properly sets universalCapable=True in build_wxwidgts.py when building on macs with only newer SDKs installed

Building universal2 wxPython required MacOS SDK  11 or newer, but only checked for SDKs up to version 12.  This patch expands to all modern SDKs which then enables proper universal2 builds.

Optionally, 1766336e555aef3b1af700f780b659e80543e2a3 can be reverted to re-enable default universal builds. [buildtools/build_wxwidgets.py#311](https://github.com/wxWidgets/Phoenix/blob/f2ef00cf13d5681ac1e06bfa9d8e7a54ecd6c644/buildtools/build_wxwidgets.py#L311) was defaulting to x86_64 when the building system had only SDKs newer than 12.0

### DIFF
--- a/buildtools/build_wxwidgets.py
+++ b/buildtools/build_wxwidgets.py
@@ -287,7 +287,7 @@ def main(wxDir, args):
             # find the newest SDK available on this system
             SDK = 'none found'
             for xcodePath in getXcodePaths():
-                possibles = [(major, minor) for major in [10, 11, 12] for minor in range(16)]
+                possibles = [(major, minor) for major in [10, 11, 12, 13, 14, 15, 26, 27] for minor in range(16)]
                 for major, minor in reversed(possibles):
                     sdk = os.path.join(xcodePath, "SDKs/MacOSX{}.{}.sdk".format(major, minor))
                     if os.path.exists(sdk):


swt2c:
Thanks, we probably could just drop support for Xcode < 11, but this is fine.

I think I'd prefer to stick with arch-specific wheels because they're smaller and we can build them in parallel (although GitHub limits the number of macOS builders anyway).  Do you have a reason we should switch back to universal?

jaknel:
A universal wheel is useful to me for distributing a mac app bundled by pyinstaller. 

Arch specific wheels require the end user to somehow install via brew or pip, an automated download at app installation time, or packaging both archs with runtime arch detection.

So I stuck with the last published universal wheel wxPython-4.2.2-cp313-cp313-macosx_10_13_universal2.whl released 2024/09/12.  This PR Is the result of building an up to date universal wheel myself.  The requirements for a functional wxPython build environment are pretty heavy, compared to the python ecosystem in general, so it would be nice to take that burden off developers who need universal builds.

edit: Would switch to one arm64 and one universal2 build be an adequate solution?  Arm64 users installing from pypi would default to the thinner build while intel users still get a functional  universal wheel that would also be useful for developers.

PR #2961, opened 2026-09-05
Use SPDX Licence Expressions / Fix setuptools install warning

swt2c:
Fixes #NNNN


echoix:
Is that LICENSE.txt committed somewhere or is still generated on the fly?

swt2c:
> Is that LICENSE.txt committed somewhere or is still generated on the fly?

Still generated on the fly.

